PR #9, opened 2021-02-09
Added access token expiration

ArnyminerZ:
Now, when an access token is generated, the generation date will be stored, and since its duration is of one hour, once that time is over, a new one will be generated. A bit of code reformatting has been done.

Also, some "security recommendations" have been added to the README.

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -5,11 +5,9 @@ See: https://gist.github.com/rithvikvibhu/952f83ea656c6782fbd0f1645059055d
 """
 import logging
 import grpc
-import json
 import datetime
 
 from gpsoauth import perform_master_login, perform_oauth
-from google.protobuf.json_format import MessageToJson
 from uuid import getnode as getmac
 
 from .google.internal.home.foyer import v1_pb2_grpc
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -6,6 +6,7 @@ See: https://gist.github.com/rithvikvibhu/952f83ea656c6782fbd0f1645059055d
 import logging
 import grpc
 import json
+import datetime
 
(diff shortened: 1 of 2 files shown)


DurgNomis-drol (approved):
Looks good to me.

This should also be applied to localAuthToken. It lives about 24 hours if i remember correctly.

ArnyminerZ:
Yeah, forgot it. Will add it ASAP

ArnyminerZ:
Added homegraph updates every 24h. Even though this is a kind of temporal fix. Error handling should be the option, I believe, but further testing must be done.

glocaltokens/client.py
```diff
@@ -86,7 +92,8 @@ def get_master_token(self):
         return self.master_token
 
     def get_access_token(self):
-        if not hasattr(self, 'access_token'):
+        if self.access_token is None or \
+                datetime.datetime.now().timestamp() - self.access_token_date.timestamp() > ACCESS_TOKEN_DURATION:
```
leikoilja, line 96: Would be nice to extract that `datetime` comparison into it's own method and reuse it

leikoilja (changes requested):
That's really nice! Thanks for the PR!

Just a minor comment to make the code a bit more reusable/testable. Will be happy to merge right when it's fixed

glocaltokens/client.py
ArnyminerZ: Makes sense, will work on it right now

leikoilja: Thank you 💥

ArnyminerZ: Added it 😉

leikoilja (approved):
Nice! Let's get it in

PR #10, opened 2021-02-10
Update explanation to make sure that others use your script corretly

DurgNomis-drol:
If you try to get a master token to often, google will disconnect your other android devices.


leikoilja (approved):
Nice! Thanks

Issue #14, opened 2021-02-11
Client's get_google_devices_json method returns list


leikoilja:
**Describe the bug**
Method that is described as `json` returns a list of devices.
See https://github.com/leikoilja/glocaltokens/blob/master/glocaltokens/client.py
```python
    def get_google_devices_json(self):
        """
        Returns a json of google devices with their
        local authentication tokens
        """

        def extract_devices(items):
            """
            Replacement for jq
            """
            devices = []
            for item in items:
                if item.local_auth_token != "":
                    device = {
                        'deviceName': item.device_name,
                        'localAuthToken': item.local_auth_token,
                    }
                    devices.append(device)
            return devices

        homegraph = self.get_homegraph()

        devices = extract_devices(homegraph.home.devices)

        return devices
```

**Expected behavior**
We should either return a json or a list. If former, then we should rename the method and announce a breaking change.

ArnyminerZ:
Implemented it in 7e1cefb and b35a8b5, but not sure why GitHub keeps adding all the commits that I have made, and not just the ones in that branch, will create a new pull request if that's not an issue, but not sure why it's happenning.

leikoilja:
Oh, that's nice, thanks @ArnyminerZ! Can you please create a new branch on your fork, then apply two of those commits on that branch and make a new pull request for these changes? Please also reference this ticket in the PR! Thanks 💥

leikoilja:
Closed in https://github.com/leikoilja/glocaltokens/commit/7b591642990635eb99dbc9e967fa1152d0eeb5bf

PR #13, opened 2021-02-10
Added device IP discovery

ArnyminerZ:
Through `pychromecast`, the IP is obtained. It's required to enter a list of accepted device model names, but I've already prepared one with the idea to add new devices, and maybe new lists, who knows, it's called `GOOGLE_HOME_MODELS`, since it contains the model names for the Google Home devices I've been able to test.

Example code:
```python
import json
from glocaltokens.client import GLocalAuthenticationTokens
from glocaltokens.scanner import ( discover_devices, GOOGLE_HOME_MODELS, )

client = GLocalAuthenticationTokens(master_token="aas_et/***")

google_devices = client.get_google_devices_json(GOOGLE_HOME_MODELS)
google_devices_str = json.dumps(google_devices, indent=2)

print('[*] Google devices', google_devices_str)
```

PR #20, opened 2021-02-13
Readme badges

ArnyminerZ:
Added some badges to the README file


leikoilja:
That looks goood! Man, we need to do something with your commits/branches history. Something went really off 😄  This PR is supposed to only add badges to README, but if you check "files changed" there are 8 files changes with plenty of changes that you have already been doing previously and should not be in this PR. 
Do you want us to schedule a call so we can look at it together and fix it up? Because as far as I see we have the exact same problem with other PR - https://github.com/leikoilja/glocaltokens/pull/13 and two of the commits that you refer here - https://github.com/leikoilja/glocaltokens/issues/14. That is also the problem why some of your commits are going off the branch and are getting lost

leikoilja:
@ArnyminerZ send me an email to `leikoilja@gmail.com` please if you are available today (Sunday, the 14th) for a short call (like 30-40 minutes) and what time would work best for you.
(i ll remove this post with my personal data after we schedule something)

Issue #17, opened 2021-02-12
Add typing


ArnyminerZ:
**Is your feature request related to a problem? Please describe.**
Modern IDEs use typing to make small errors like typos, or bad types mixing easier with typing. This should be implemented to have a cleaner code, since the variable types would be known.

**Describe the solution you'd like**
Add typing to all the expressions in the library, for example:
```python
def __init__(self, username:str = None, password:str = None):
```

leikoilja:
@ArnyminerZ, thanks for the issue. Do you have an interest to take a lead on this one?

ArnyminerZ:
I will take it ☺️

PR #22, opened 2021-02-15
Implement typing

ArnyminerZ:
I've implemented typing on I believe all the functions. I've also added some tests for token format checking, and variable type checks.

Closes https://github.com/leikoilja/glocaltokens/issues/17


leikoilja (commented):
Looks really good, thanks for implementing it! 💥 

Let's wait to merge the https://github.com/leikoilja/glocaltokens/pull/13 then rebase to latest `master` on this repo and then the only changes left here will be from implementing typing and type checks. It will be easier to review and I ll do it right away, so we can get it merged 👍

leikoilja:
Now that the #13 is merged, we have got some conflicts here. Can you please rebase it to the latest `master` and force push your changes into this branch? then i'll be able to review it easier 😄

ArnyminerZ:
I believe all the conflicts should be fixed

glocaltokens/client.py
```diff
@@ -35,9 +37,72 @@
...
+        else:
+            if (ip and not port) or (not ip and port):
+                LOGGER.error(
+                    "Both ip and port must be set, if one of them is specified."
+                )
+                return
+            self.ip = ip
+            self.port = port
+
+        if self.ip and not net_utils.is_valid_ipv4_address(self.ip) and not net_utils.is_valid_ipv4_address(self.ip):
```
leikoilja, line 80: The typo here again with ipv4 and ipb6

glocaltokens/client.py
```diff
@@ -35,9 +37,72 @@
...
+            return
+
+        if self.port and not net_utils.is_valid_port(self.port):
+            LOGGER.error("port must be a valid port")
+            return
+
+    def __str__(self) -> str:
+        return str(self.dict())
+
+    def dict(self) -> []:
```
leikoilja, line 91: isn't the return type here `Dict`?

glocaltokens/client.py
```diff
@@ -163,49 +233,49 @@ def get_google_devices(self, models_list=None):
...
         :param models_list The list of accepted model names.
         :param indent The indentation for the json formatting.
         """
-        devices = self.get_google_devices(models_list)
-        return json.dumps(devices, indent=indent)
+        
+        devices_json = [
+            device.dict() for device in self.get_google_devices(models_list)
+        ]
+        return json.dumps(devices_json, indent=indent)
```
leikoilja, line 281: Would be great to add a Client test for that one. This method must be quite robust, so it's important we have it covered with tests. Will create a new issue for that

glocaltokens/scanner.py
```diff
@@ -91,8 +93,8 @@ def get_value(key):
 
 
 class GoogleDevice:
-    def __init__(self, name, ip, port, model):
-        if not net_utils.is_valid_ipv4_address(ip) and not net_utils.is_valid_ipv6_address(ip):
+    def __init__(self, name: str, ip: str, port: int, model: str):
+        if not net_utils.is_valid_ipv4_address(ip) and not net_utils.is_valid_ipv4_address(ip):
```
leikoilja, line 97: Why do you change it to ipv4 here?

tests/factory/providers.py
```diff
@@ -0,0 +1,14 @@
...
+import string
+from faker import Faker
+from faker.providers import BaseProvider
+
+fake = Faker()
+
+
+class TokenProvider(BaseProvider):
+    def aas_et(self):
+        return "aas_et/" + "".join(random.choice(string.ascii_letters) for x in range(216))
```
leikoilja, line 11: Let's reuse the method from `glocaltokens/utils/token.py` here, so we don't duplicate them

tests/factory/providers.py
```diff
@@ -0,0 +1,14 @@
...
+
+    def local_auth_token(self):
+        return "".join(random.choice(string.ascii_letters) for x in range(108))
```
leikoilja, line 14: Same here

tests/test_client.py
```diff
@@ -20,7 +29,7 @@ def tearDown(self):
     def test_initialization(self):
         username = faker.word()
         password = faker.word()
-        master_token = faker.word()
+        master_token = faker.aas_et()
```
leikoilja, line 32: can we maybe prefix the `aas_et` with `token_aae_et`? It would be more elaborate then and we could introduce easily another token types if needed

tests/test_client.py
```diff
@@ -34,25 +43,41 @@ def test_initialization(self):
         self.assertEqual(master_token, client.master_token)
         self.assertEqual(android_id, client.android_id)
 
+        self.assertTrue(type(client.username) == str)
```
leikoilja, line 46: I suggest we do a separate test mixin for that type of checks. It would look like this:

```python
class TypeTestMixin:
 
  def assertIsString(self, variable):
    self.assertTrue(type(variable) == str, msg=f"Given variable {variable} is not String type")

class GLocalAuthenticationTokensClientTests(TypeTestMixin, TestCase):
   ...
   self.assertIsString(client.master_token)
```

leikoilja (changes requested):
Awesome stuff! Just a few small comments

glocaltokens/client.py
ArnyminerZ: Yeah, the issue is that I didn't know how to test it, since it's calling methods that require your Google Account, how to I build tests on that one?

glocaltokens/client.py
ArnyminerZ: Fixed 29f19f4

glocaltokens/client.py
ArnyminerZ: Yeah, I don't know why I put that. Fixed 6155351.

glocaltokens/scanner.py
ArnyminerZ: 🤔 8732573

tests/factory/providers.py
ArnyminerZ: Unified in a function which may also be useful in the future. c2d5516

tests/factory/providers.py
ArnyminerZ: Check the [other message](https://github.com/leikoilja/glocaltokens/pull/22#discussion_r576667308).

tests/test_client.py
ArnyminerZ: Renamed to `token_aas_et`. 7d7fa1d

tests/test_client.py
ArnyminerZ: Added `assertIsString` and `assertIsAASET` in 52ea81d

glocaltokens/client.py
leikoilja: We can use `mock.patch` for this.
You can leave it for me, I can do that after we merge this one (because it will rely on token generation) and then you can take a look of how to do that and maybe learn something new from it :)

tests/test_client.py
leikoilja: @ArnyminerZ it i's great that you are refering to specific commits everytime you fix something, but if it takes too much time you can skip doing that. If the conversation is ongoing just keep it "unresolved" and if you agree with a comment and fixed something you can simply press "resolve conversation". Because when you re-request the review, I will still go through all the files again and see all changes at once :)

leikoilja:
@ArnyminerZ, we have a new cool `pre-commit` github action that does a lot of code linting and checks so we have the code consistent. You can read more about it here: https://pre-commit.com/

But all in all, what you need to do is install `pip pre-commit` first and then run `pre-commit install` in the working directory. This will install `git` hooks that will be run EVERY time you commit the code. It will happen automatically, but for now just run `pre-commit run --all-files` to manually trigger it and you will see a lot of files being changed and linted :) 
After lining, commit the code and push it here. You will see the `pre-commit` action pass

tests/test_client.py
ArnyminerZ: Ok, I don't mind referencing the issues, it's a matter of seconds, but if it's not useful, I will just mark them as resolved 😉

tests/factory/mixin.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class TypeTestMixin:
+
+    def assertIsString(self, variable):
+        self.assertTrue(type(variable) == str, msg=f"Given variable {variable} is not String type")
+
+    def assertIsAASET(self, variable):
+        self.assertTrue(type(variable) == str and token_utils.is_aas_et(variable),
+                        msg=f"Given variable {variable} doesn't follow the AAS_ET format")
```
leikoilja, line 11: Awesome stuff adding it in a factory directory. It's not really a factory at this point, but that's just details, let's keep it here for now :)

tests/factory/providers.py
```diff
@@ -0,0 +1,13 @@
...
+
+fake = Faker()
+
+
+class TokenProvider(BaseProvider):
+    def token_aas_et(self):
+        return generate_token(216, prefix="aas_et/")
+
+    def local_auth_token(self):
+        return generate_token(108)
```
leikoilja, line 13: Awesome stuff

tests/test_client.py
```diff
@@ -37,25 +46,41 @@ def test_initialization(self):
...
+
         self.assertIsNone(client.access_token)
         self.assertIsNone(client.homegraph)
         self.assertIsNone(client.access_token_date)
         self.assertIsNone(client.homegraph_date)
 
+        self.assertIsAASET(client.master_token)
+
+        # Test get_android_id
+        client = GLocalAuthenticationTokens(username, password)
```
leikoilja, line 62: You already have a client here, no need to create a new one, just use the existing on

tests/test_client.py
```diff
@@ -37,25 +46,41 @@ def test_initialization(self):
...
+        android_id = client.get_android_id()
+        self.assertIsNotNone(android_id)
+        self.assertIsString(client.android_id)
```
leikoilja, line 65: You can even move this check into the test we already have `test_get_android_id`

leikoilja (changes requested):
Great stuff! Please see a few comments and make sure to run `pre-commit` hooks to lint the code using Black standard.

See my previous comment about `pre-commit`

leikoilja (approved):
Awesome work! Thank you.

Let's see the tests pass and then merge it in! 💥

Issue #11, opened 2021-02-10
Package wheels fails to build because of grpcio, grpcio-tools missing system packages/compilers


leikoilja:
**Describe the bug**
Package fails to build because it relies on wheels for `grpcio`, `grpcio-tools`. 

**Expected behavior**
It would be nice to find a solution in-line with current [Home Assistant](https://github.com/home-assistant/core/) version, so the package would work with HA on arm devices too. Preferably without installing any additional system packages/compilers, since not every user will have access/knowledge to do so.

Reference from https://github.com/leikoilja/glocaltokens/issues/6#issue-799619106: 
> But i ran into this problem with the docker version of homeassistant, because it runs on an alpine image, grpcio and grpcio-tools dosent install corretly if you dont install them manually in the docker image, and add "apk add gcc g++ linux-headers".

leikoilja:
Oh, we had the wrong docker image in the `devcontainer` in this repo. It was set to `debian`, so the building of grpcio was failing. I have changed it to `alpine` so it will  correspond to RPi container now. Plus will add a comment from @DurgNomis-drol  to run `apk add gcc g++ linux-headers` before installing the integration to the documentation. Also will keep the `grpcio` version pinned to `grpcio==1.31.0` for the `glocaltokens` as is it specified in HomeAssistant core repo

DurgNomis-drol:
I just tested it and we only need `g++` and `linux-headers` to make it able to build on alpine. `gcc` is installed when you install `g++`.

leikoilja:
That's great, thank you @DurgNomis-drol. I am testing now with just `linux-headers` on alpine image as well, will see if that can solve it. Either way, as we discussed in https://github.com/leikoilja/ha-glocaltokens/pull/21#issuecomment-778581649 once we know which library/libraries are essential to build `grpcio` wheels I will try opening issue/PR with HA to have those libs included in the HA build

leikoilja:
From HomeAssistant core [repo](https://github.com/home-assistant/core/blob/6e67b943da09a13409e1a61ec34ed7144e1093c3/script/gen_requirements_all.py)

> gRPC 1.32+ currently causes issues on ARMv7, see:
> https://github.com/home-assistant/core/issues/40148
>

DurgNomis-drol:
I have no problem building new versions on arm if i add `g++` and `linux-headers` and the install `grpcio` and `grpcio-tools` manually

leikoilja:
[...]

PR #29, opened 2021-02-17
Grpcio tools version

ArnyminerZ:
I've had problems installing `grpcio` on some devices because not setting a version on `grpcio-tools` installs the latest one, which is incompatible with `grpcio==1.31.0`. Adding this should fix this, and it's not a big change.

*Note: I can't figure out why there are that much commits, but it only updates one file, so I think it doesn't matter.*


DurgNomis-drol:
This issue is know already

And `grpcio-tools` still need `g++` and `linux-headers` to compile on alpine. Which is not in the official release of HA.

ArnyminerZ:
> And `grpcio-tools` still need `g++` and `linux-headers` to compile on alpine. Which is not in the official release of HA.

And isn't there a way to install those packages when the library is installed, maybe through `setup.py`? 🤔

DurgNomis-drol:
As far as i know there is not.
I don't think that HA allows integrations it install non python dependencies.

It have to be done by HA when the alpine image is build.

DurgNomis-drol:
See #11 for more details

DurgNomis-drol:
And [Here](https://github.com/leikoilja/ha-glocaltokens/issues/14)

ArnyminerZ:
Yes, the long compilation times for RPi are a bit unevitable, since it's a quite low-power computer... Even that, what I meant is that python installs `grpcio-tools` version `1.35.0` if version is not set. This version is incompatible with `grpcio` version `1.31.0`, which is the required by `glocaltokens`, and it's required to install `grpcio-tools==1.31.0` manually.

DurgNomis-drol:
This is already know, but this problem has to be fixed on HA side. This will also remove the issue of long build times for `grpio-tools` on RPI. Because `grpcio` and `grpcio-tools` will then have precompiled build wheels from HA. And this problem also exists on other systems then RPI if they run on alpine. HomeassistantOS is based on buildroot and not Alpine.

But yes `grpcio-tools` should be changed to `grpcio-tools` version that matches the already installed version of `grpcio` in HA.

I don't not know the reason why HA does not ship with latest version of `grpcio`.

@leikoilja Have already opened an issue over at HA to fix these things. Until then, you would have to install it manually.

Also you have to fix your commit to only include the change of dependency for `grpcio.tools` :smile:

leikoilja (approved):
That's a great find, thank you, @ArnyminerZ.

leikoilja:
It is indeed an issue that we need to manually build `grpcio` and `grpcio-tools` outside the home assistant. That takes time and is inconvenient for most users, but before HA solves that we'll have to just manually install those. If I understood it correcntly they have locked the version down for a reason, i believe the newest versions were failing to build on `arm`:

> 
> gRPC 1.32+ currently causes issues on ARMv7, see:
> https://github.com/home-assistant/core/issues/40148

Here is a general issue where we can keep the discussion regarding `grpcio` - https://github.com/leikoilja/glocaltokens/issues/11

Issue #23, opened 2021-02-15
Add tests for device json fetching


leikoilja:
**Describe the bug**
Would be great to add a Client test for that one. This method must be quite robust, so it's important we have it covered with tests. 

**Expected behavior**
Nice and clean tests to ensure we have got that functionality covered

Blocked by #22

ArnyminerZ:
Closed by #31

PR #31, opened 2021-02-18
Get devices testing

ArnyminerZ:
Added `get_google_devices` and `get_google_devices_json` checks, as well as some optimizations.

Aims to #23


ArnyminerZ:
@leikoilja please, can you take a look at the tests? Not sure what am I doing wrong here.

glocaltokens/client.py
```diff
@@ -203,61 +202,66 @@ def get_homegraph(self):
...
-    def get_google_devices(self, models_list=None) -> [Device]:
+    def get_google_devices(
+        self, models_list: Optional[List[str]] = None, disable_discovery: bool = False
+    ) -> [Device]:
         """
         Returns a list of google devices with their local authentication tokens, and IP and ports if set in models_list.
 
-        :param models_list The list of accepted model names.
+        models_list: The list of accepted model names.
+        disable_discovery: Whether or not the device's IP and port should be searched for in the network.
```
leikoilja, line 212: Great stuff!

glocaltokens/client.py
```diff
@@ -203,61 +202,66 @@ def get_homegraph(self):
...
+        indent: int = 2,
+        disable_discovery: bool = False,
     ) -> str:
         """
         Returns a json list of google devices with their local authentication tokens, and IP and ports if set in
         models_list.
 
-        :param models_list The list of accepted model names.
-        :param indent The indentation for the json formatting.
+        models_list The list of accepted model names.
```
leikoilja, line 256: Maybe a `:` sign after parameters, like `models_list:`

glocaltokens/const.py
```diff
@@ -8,3 +8,11 @@
...
 
 HOMEGRAPH_DURATION = 24 * 60 * 60
+
+DISCOVER_TIMEOUT = 20
+GOOGLE_HOME_MODELS = [
+    "Google Home",
+    "Google Home Mini",
+    "Google Nest Mini",
+    "Lenovo Smart Clock",
+]
```
leikoilja, line 18: Good call to move it into `const`

leikoilja (changes requested):
Those are some really nice refactoring! 

Uh that's actually a tough tests to make, now that i am thinking about it. 
To fully test it, we will probably need to  patch `homegraph.home.devices` to return a list of objects like homegraph would actually return (just with a fake data). And then validate that we serialize and create `Device` objects correctly from that received payload.
I will try looking at it later tonight to see how can I help with that :) 

We could also create a separate label for `tests` if you feel it might be useful! Feel free to create it, you should have permissions to do so 👍

ArnyminerZ:
> To fully test it, we will probably need to patch `homegraph.home.devices` to return a list of objects like homegraph would actually return (just with a fake data). And then validate that we serialize and create `Device` objects correctly from that received payload.

Maybe creating a provider for fake Homegraph device data should do the trick? Just like we do with tokens.

ArnyminerZ:
Not sure why this isn't working on the json tests, but it does on the normal one. Will keep researching... Pushing here just for reference and progress updates.

ArnyminerZ:
Found out that for some reason when `get_google_devices_json` is called from `test_google_devices_json`, calls `self.get_google_devices` in some way that makes it return something like `<MagicMock name='get_google_devices()' id='140244469296240'>`, which isn't an array, what makes the loop not to run, and return an empty list.

leikoilja:
Lemme take a looook, @ArnyminerZ  👀 😄  Will push some change if/when i find something

leikoilja:
@ArnyminerZ, please see some of the changes I propose. It might not be the cleanest or the best way to do it, but maybe it can be a starting point?
image

ArnyminerZ (commented):
Awesome work! Everything seems right to me.

ArnyminerZ:
> 
> 
> @ArnyminerZ, please see some of the changes I propose. It might not be the cleanest or the best way to do it, but maybe it can be a starting point?

Yeah it's a bit dirty, but if for now it works, it's fine for me. In the future we may improve this.

tests/factory/const.py
```diff
@@ -0,0 +1,173 @@
+# Right now this is pretty useless, a great confusion. Don't delete just in case it's useful in the future.
```
leikoilja, line 1: Hm... this is a good find, might be useful sometime down the road. Shall we just move it to a separate branch, commit and leave it there for the time being?

tests/factory/providers.py
```diff
@@ -1,17 +1,107 @@
 from faker import Faker
 from faker.providers import BaseProvider
 
+from glocaltokens.const import (
+    ACCESS_TOKEN_LENGTH,
+    LOCAL_AUTH_TOKEN_LENGTH,
+    MASTER_TOKEN_LENGTH,
+)
```
leikoilja, line 8: Thanks for doing it, great idea!

tests/factory/providers.py
```diff
@@ -1,17 +1,107 @@
...
-        return generate_token(315, prefix="ya29.")
+        return generate_token(ACCESS_TOKEN_LENGTH, prefix="ya29.")
 
     def local_auth_token(self):
-        return generate_token(108)
+        return generate_token(LOCAL_AUTH_TOKEN_LENGTH)
+
+
+class HomegraphProvider(BaseProvider):
+    # TODO: This should be moved somewhere else
```
leikoilja, line 29: Do we need this comment this? Did you refer to the faker of `version`?

tests/factory/providers.py
```diff
@@ -1,17 +1,107 @@
...
+    def version(self):
+        return str(fake.pyint()) + "." + str(fake.pyint()) + "." + str(fake.pyint())
```
leikoilja, line 31: Can we not do it in a bit more pythonic way? 😄 Maybe like so:
`"{}.{}.{}".format(fake.pyint(), fake.pyint(), fake.pyint())`
or f-strings
`f"{fake.pyint()}{fake.pyint()}{fake.pyint()}"`

Although i live the `format` method more, seems more visual to me 😄

tests/factory/providers.py
```diff
@@ -1,17 +1,107 @@
...
+
+    def google_device(self):
+        """
+        Not sure from where did I get this structure from. Just using the content from client.py as reference.
+        """
```
leikoilja, line 36: We can just keep a shorted comment here like: "Using homegraph returned data as a reference" or smth similar

tests/factory/providers.py
```diff
@@ -1,17 +1,107 @@
...
+    #      """
+    #      Generates a random amount of devices, in the range specified.
+    #
+    #      min_devices: The number minimum of devices to generate. Should be greater than 0
+    #      max_devices: The maximum number of devices to generate. Must be greater than min_devices
+    #      """
+    #      return [
+    #          self.homegraph_device()
+    #          for n in range(fake.random.randint(min_devices, max_devices))
+    #      ]
```
leikoilja, line 107: Do you think we still need it? And if so, what is the difference between homegraph_device and google_device?

leikoilja (approved):
I have a few questions regarding comments and cleaning up, but in general I agree with most of it! 💥  I'm going to approve, but please take a look on the comments if we can clean it up a bit before merging-in? 👍 
When you feel satisfied - merge it in 💥

tests/factory/providers.py
ArnyminerZ: Ok, that's fine, it was just to inform you about that.

tests/factory/providers.py
ArnyminerZ: Yes, I mean moving the version generator to a new provider since it's not properly related to Homegraph.

tests/factory/providers.py
ArnyminerZ: `homegraph_device` is the model that is providen by the Google Documentation. I don't remember where did I get the `google_device` model from, but at the end, it's what we get from the RPC request. We might not need this, but it's the same as the content at `const.py`, it might be helpful, but not for now. As you've said, it may be appropiate to move it somewhere else.

tests/factory/providers.py
leikoilja: If so, let's just create a custom Provider like `UtilsProvider` or smth, where we can move/add all similar generic stuff?

tests/factory/providers.py
leikoilja: How about we move both to a separate branch `homegraph_mock` or smth like that on this repo?

leikoilja (approved):
Great stuff, thank you! 

Just a few minor comments and then feel free to merge

tests/factory/providers.py
ArnyminerZ: Ok, sounds fair, done that

ArnyminerZ:
@leikoilja why is the `Labeler` failing? 🤔 Is that important?

leikoilja:
@ArnyminerZ, no no, that's not important yet. I will try fixing it today, not exactly sure why it doesn't have permissions to edit labels. It's a new github action that will help us automatically assign labels on PRs depending on what files were changes. For example, when `tests/*` changed it will assign `testing` label. See this config: [here](https://github.com/leikoilja/glocaltokens/blob/master/.github/labeler.yml)

ArnyminerZ:
Then I will merge once `3.9` test is complete. Even though it should pass without any issues.

PR #37, opened 2021-03-05
Fix authentication (401) error in `get_homegraph`

### LINKED ISSUE #36: Authentication error after ~22h of usage
**Describe the bug**
Left the `Google Home` HA integration for 24h and after the time has passed I have discovered that integration fails to fetch data due to authentication error.
image

While this should not happen and tokens must be automatically refreshed by `glocaltokens` library. I am afraid it might not be working... OR we might need to shorten the access token and homegraph durations. @ArnyminerZ, do you have any thoughts?

ArnyminerZ:
Should fix #36

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -195,16 +195,26 @@ class GLocalAuthenticationTokens:
         if self.homegraph is None or self._has_expired(
             self.homegraph_date, HOMEGRAPH_DURATION
         ):
-            scc = grpc.ssl_channel_credentials(root_certificates=None)
-            tok = grpc.access_token_call_credentials(self.get_access_token())
-            ccc = grpc.composite_channel_credentials(scc, tok)
+            try:
+                scc = grpc.ssl_channel_credentials(root_certificates=None)
+                tok = grpc.access_token_call_credentials(self.get_access_token())
+                ccc = grpc.composite_channel_credentials(scc, tok)
 
-            with grpc.secure_channel(GOOGLE_HOME_FOYER_API, ccc) as channel:
-                rpc_service = v1_pb2_grpc.StructuresServiceStub(channel)
-                request = v1_pb2.GetHomeGraphRequest(string1="", num2="")
-                response = rpc_service.GetHomeGraph(request)
-                self.homegraph = response
-            self.homegraph_date = datetime.now()
+                with grpc.secure_channel(GOOGLE_HOME_FOYER_API, ccc) as channel:
+                    rpc_service = v1_pb2_grpc.StructuresServiceStub(channel)
+                    request = v1_pb2.GetHomeGraphRequest(string1="", num2="")
+                    response = rpc_service.GetHomeGraph(request)
+                    self.homegraph = response
+                self.homegraph_date = datetime.now()
+            except grpc.RpcError as rpc_error:
+                if rpc_error.code() == 401:
+                    LOGGER.warning("The access token has expired. Getting a new one.")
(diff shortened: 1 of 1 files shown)


glocaltokens/client.py
```diff
@@ -195,16 +195,26 @@ def get_homegraph(self):
...
+
+                with grpc.secure_channel(GOOGLE_HOME_FOYER_API, ccc) as channel:
+                    rpc_service = v1_pb2_grpc.StructuresServiceStub(channel)
+                    request = v1_pb2.GetHomeGraphRequest(string1="", num2="")
+                    response = rpc_service.GetHomeGraph(request)
+                    self.homegraph = response
+                self.homegraph_date = datetime.now()
+            except grpc.RpcError as rpc_error:
+                if rpc_error.code() == 401:
+                    LOGGER.warning("The access token has expired. Getting a new one.")
```
DurgNomis-drol, line 211: This only fixes it if the access token has expired, not if the local auth token has.

leikoilja:
@ArnyminerZ, I have been thinking about it and i'm afraid it might not be fixing out local token authentication issue, just as @DurgNomis-drol mentioned. Sorry for confusion 😄  (i'll try explain better in the issue)

We should keep this fix because it is really important to cover access token failure as you do here 💯  It would be super good to have some tests on that but it might be hard to make them haha 😄

glocaltokens/client.py
ArnyminerZ: It should refresh the local auth token as well, since it's loaded from there. See [here](https://github.com/leikoilja/glocaltokens/blob/master/glocaltokens/client.py#L241)

leikoilja:
@ArnyminerZ, can you think of any good way we can test it before releasing? 😄

glocaltokens/client.py
DurgNomis-drol: I have been thinking about and you are right, i was getting confused. I thought that we stored the token in ha-google-home, but we don't, we asks for it every time. Sorry for wasting your time 😄

ArnyminerZ:
> 
> 
> @ArnyminerZ, can you think of any good way we can test it before releasing? 😄

Well, I think that passing a virtual auth token to `get_homegraph` and trying to fetch from the server should return 401, but then solving the problem... It not clear, because it'd try to get a new token with invalid credentials, and blah, blah, blah.
So... It's quite a problem to test this... 🤔

leikoilja:
Sounds gooood! I'll try putting a debugger there and fake token to see 401 returned and see if it will return the right token back....

leikoilja:
Wohooo, the stuff in amazing, thanks @ArnyminerZ! 💥  🥳 
I will try to test it manually in a bit and if all works - let's merge it in!

leikoilja:
It actually didn't work. Providing the wrong access token result in log error:

> ERROR:glocaltokens.client:Received unknown RPC error: code=StatusCode.UNAUTHENTICATED message=Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or othe
> r valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.
> 

Apparently, it doesn't have 403 code there anymore. They have a value of it `16` which is probs their internal reference. So i have modified the check a bit and tested it to work :)

leikoilja (approved):
Thank you! Feel free to merge, if you agree with my changes

ArnyminerZ:
Just made a guess there, that's what tests are for 😄 
Squashing 🔨  and merging 🚀

PR #38, opened 2021-03-05
Expose hardware name in the Device

leikoilja:
I have a problem in ha-google-home that it tries fetching alarms and timers from chromecast devices, where those don't even exist. Those can be simply fixed if we know the hardware model of the devices and only create timer/alarm sensors for devices that are supported like "Google Home", "Google Home Mini" etc


ArnyminerZ (approved):
Seems good to me

PR #39, opened 2021-03-06
Making it possible to store an android_id and reuse it later

DurgNomis-drol:
Making it possible to store an android_id and reuse it later, so we don't generate a new android every time we call `get_google_devices` or `get_google_devices_json`.

This is a simple fix, if some one have a better one. Just close this and PR the better one. 😄

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -227,6 +227,7 @@ class GLocalAuthenticationTokens:
         disable_discovery: bool = False,
         zeroconf_instance=None,
         force_homegraph_reload: bool = False,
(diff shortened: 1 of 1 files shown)


leikoilja:
Thank you, @DurgNomis-drol for taking a lead on this one! 
There was my confusion in `ha-google-home`, sorry 😄  
Also, if you would need android_id in the `get_google_devices` method you can always access it from class variables like `self.android_id` or even better `get_android_id()`. And if you would need to set that variable you could do it from `ha-google-home` during initialization or by saying `client.android_id = 'your text'` where `client = GlocaltokensClient` 👍

PR #41, opened 2021-03-08
Adding CI using Github workflows

samapriya:
Adding github workflows for CI for glocaltokens

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,7 @@
 # This workflow will install Python dependencies, run tests and lint with a variety of Python versions
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
 
-name: CI glocaltokens
+name: Build package
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


KapJI:
I'd name it `build.yaml`. And set name to something like "Build package".

leikoilja:
Awesome, thanks @samapriya!
Haha, how come it passes windows build when it was failing for you locally, @samapriya ? 🤔 
@KapJI is saying some good advice! Could you please rename the workflow file and it's label? :)

samapriya:
Seems it is automatically reverting to v0.2.3 for windows and the installer still installs the dependencies even if v0.2.4 fails, I was confused too so I had to look at the logs to see that the installed version on windows goes back to 0.2.3 once it fails for 0.2.4

KapJI:
I can't see where it fails on Windows from logs. Also `setup.py` still has `0.2.2` version but it should use code from the master anyway.

samapriya:
My bad yeah I just found the setup.py having 0.2.2 as well , I think I was testing my initial setup with a sneaky pip install glocaltokens initally which I changed to python setup.py install. I think the setup.py version should be updated and that might trigger it, checking now after updating setup.py version

samapriya:
Am not a 100% sure anymore because checks seem to be passing with v0.2.4 on windows too. My best guess here is if you run a CI with python setup.py install it seems to working but a pip install glocaltokens is failing. So this is an issue with packaging and should be resolved if the history object is removed. You can try this by modifying the .yml file to do a pip install glocaltokens instead a python setup.py install

samapriya:
Tested it on my own branch check the results here
https://github.com/samapriya/glocaltokens/pull/1/checks?check_run_id=2058632090

PR #42, opened 2021-03-12
Enable distributing type info with a package

leikoilja:
To make the package [PEP-561](https://www.python.org/dev/peps/pep-0561/) compliant this PR 
enables distributing typing info
 
Required by: https://github.com/leikoilja/ha-google-home/pull/88

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,7 @@ classifiers = [
     "License :: OSI Approved :: MIT License",
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
+    "Typing :: Typed",
 ]
 description = "Tool to extract Google device local authentication tokens in Python"
 keywords = [
@@ -15,6 +16,7 @@ keywords = [
     "Google",
     "LocalAuthenticationTokens",
 ]
+include = ["glocaltokens/py.typed"]
 license = "MIT"
 readme = "README.md"
 homepage = "https://github.com/leikoilja/glocaltokens"


leikoilja:
@KapJI, please take a look

KapJI:
You also need to create empty `py.typed` file in package root.

KapJI:
Also do you have `mypy` in CI jobs?

leikoilja:
No, there is no `mypy` check in CI jobs. I ll add a separate issue to introduce `poetry` in this repo and enable `mypy` correspondingly to what you have done in `ha-google-home` :)

KapJI (approved):
Make sure you run `mypy` before release the new package, otherwise errors can be propogated to packages which use `glocaltoken`.

leikoilja:
hmmm, why would that be a reason, @KapJI? You mean `typing` errors? They shouldn't break anything, right?

KapJI:
They can break type checking. Basically `mypy` will check all dendencies which are marked as typed with `py.typed`.

KapJI:
@leikoilja this seems stuck. Is there anything blocking you?

leikoilja:
yeh, took a little break, but now i m back. Also had started on setting up `poetry` for this repo and it went down with TONS of errors. Couldn't really understand whether it's because of my experimental `arm m1 mac` that has some glitches here and there or something else. WIll pick it up soon on my ubuntu and proceed with `poetry` :)

leikoilja:
@KapJI can't find any useful information in docs, but now that the `poetry` is in place, do we need to explicitly note something like `package_data={"glocaltokens": ["py.typed"]}` that used to be in `setup.py`?

KapJI:
Yes, you can do this using `include`: [docs](https://python-poetry.org/docs/pyproject/#include-and-exclude).
Also add `Typing :: Typed` classifier.

KapJI (approved):
Yeah, should be it :)

PR #45, opened 2021-03-13
Don't store Device if the local_auth_token is invalid

### LINKED ISSUE #44: Do not initialise Devices with invalid `local_auth_token`
**Describe the bug**
When we get a fresh `homegraph` we iterate through devices creating new instances `Device` [ref](https://github.com/leikoilja/glocaltokens/blob/e6fa89793bc72c84b3b61f99c371fdb6d7ba8c2f/glocaltokens/client.py#L271) for each of discovered devices. In the `Device` initialization `__init__` we do the validation for `local_auth_token`, however, if the validation fails we still return the "half-baked" object back to `devices` that is later used.

** Referece **
See the issue [here](https://github.com/leikoilja/ha-google-home/issues/91)

**To Reproduce**
Put a debugger breakpoint right before token validation and temper with a token. 

**Expected behavior**
If any of the validations fail - skip adding the device

ArnyminerZ:
When initializing `Device`, `local_auth_token` won't get stored until all the checks are passed. This is used in `get_google_devices` for checking if the class has been initialized correctly, and if not, don't adding it to the devices list.

If the device is not added, a warning is shown to through the `LOGGER`.

*Note: Maybe a more detailed debug would be convenient*


glocaltokens/client.py
```diff
@@ -52,15 +52,15 @@ def __init__(
...
             return
 
         if not self.device_name:
             LOGGER.error("device_name not set")
             return
 
-        if not token_utils.is_local_auth_token(self.local_auth_token):
+        if not token_utils.is_local_auth_token(local_auth_token):
             LOGGER.error("local_auth_token doesn't follow the correct format.")
             return
```
leikoilja, line 65: Sorry for being a bit picky, Can we move all token related validation together, so `device_name` check is no in-between 😄

glocaltokens/client.py
```diff
@@ -88,6 +88,8 @@ def __init__(
             LOGGER.error("port must be a valid port")
             return
 
+        self.local_auth_token = local_auth_token
```
leikoilja, line 91: Line 51 already does that assignment. Maybe we should start the `__init__` with:
`self.local_auth_token = None` as default and then do all the validation and if all the validations pass assign the token `self.local_auth_token = local_auth_token`? :)

glocaltokens/client.py
```diff
@@ -52,15 +52,15 @@ def __init__(
         self.google_device = google_device
```
KapJI, line 52: It's still set here, is it correct?

glocaltokens/client.py
```diff
@@ -267,14 +269,16 @@ def find_device(name) -> Optional[GoogleDevice]:
...
-                        hardware=item.hardware.model,
-                    )
+                device = Device(
+                    device_name=item.device_name,
+                    local_auth_token=item.local_auth_token,
+                    google_device=google_device,
+                    hardware=item.hardware.model,
                 )
+                if device.local_auth_token != item.local_auth_token:
+                    LOGGER.warning("Device initialization failed, skipping.")
```
leikoilja, line 279: This is an explicit check, you would be just fine by saying 
`if device.local_auth_token` because as you have just made it - it will only be initialized if the token is passed and is valid. But it's totally up to you what style you prefer :)

glocaltokens/client.py
KapJI: It should be probably set to `None`.

leikoilja (changes requested):
Thank you so much, @ArnyminerZ! 💥  You are the ⭐ !

Now that this patch is fixing a bug, we should really introduce a test for that, to ensure the same bug would never happen again.
There is a good methodology for bug solving: 
- if a bug occurs - find the bug
- write a test that should replicate the bug (the test must fail at this point)
- fix the bug in the code
- re-run the test (it must pass now) 

It would be amazing if you would have time to try knocking off the test for this one! If not, lemme know, i ll find some time to add it :)

glocaltokens/client.py
ArnyminerZ: If doesn't matter, because it's checked that `self.ip` and `self.port` are valid IPs afterwards, so in case `google_device` is `None`, or `ip` or `port` are null, the function will be returning, and it will detect "an error".

See line 67

ArnyminerZ:
@leikoilja take a look at the tests, not sure if that was what you meant.

glocaltokens/client.py
ArnyminerZ: That was totally a typo. Fixed it

glocaltokens/client.py
KapJI: Sorry, I meant that `self.local_auth_token` shouldn't be set twice and constructor should always set the same number of attributes. Github doesn't allow to add comment to line 51 here.

glocaltokens/client.py
ArnyminerZ: Yeah, I like more that. Quite cleaner.

glocaltokens/client.py
ArnyminerZ: Not sure if this was what you meant, take a look

ArnyminerZ:
`test_get_google_devices` is now not initializing `google_devices` well, working on it

ArnyminerZ:
Not sure what's going wrong... @leikoilja @KapJI can you help me?

glocaltokens/client.py
```diff
@@ -267,14 +269,16 @@ def find_device(name) -> Optional[GoogleDevice]:
...
+                device = Device(
+                    device_name=item.device_name,
+                    local_auth_token=item.local_auth_token,
+                    google_device=google_device,
+                    hardware=item.hardware.model,
                 )
+                if device.local_auth_token:
+                    LOGGER.warning("Device initialization failed, skipping.")
+                else:
+                    devices.append(device)
```
KapJI, line 281: Well, fake device in test is now filtered out here.

leikoilja: Indeed, you probably want the logic reversed:
`if not device.local_auth_token:`

ArnyminerZ: Totally my fault 🤦🏼 That's what tests are for 😅

ArnyminerZ (commented):
Added changes

tests/test_client.py
```diff
@@ -284,7 +284,7 @@ def test_get_google_devices(self, m_get_homegraph):
         self.assertEqual(google_device.hardware, homegraph_device.hardware.model)
 
         # With many devices returned from homegraph
-        homegraph_devices = faker.google_devices()
+        homegraph_devices = faker.homegraph_devices()
         m_get_homegraph.return_value.home.devices = homegraph_devices
 
         # With no discover_devices, with no model_list
```
leikoilja, line 290: We would still need to add after this part one more clause something like (untested code):
```python
# With many devices returned from homegraph
# but one device having the invalid token
homegraph_device_valid = faker.homegraph_devices(count=1)[0]  # TODO: we probs need `faker.homegraph_device()` method
homegraph_device_invalid = faker.homegraph_devices(count=1)[0]
homegraph_device_invalid.local_auth_token = faker.word(). # setting invalid token intentionally
m_get_homegraph.return_value.home.devices = [homegraph_device_valid, homegraph_device_invalid]
google_devices = self.client.get_google_devices(disable_discovery=True)
self.assertEqual(len(homegraph_devices), 1)
self.assertEqual(homegraph_devices[0], homegraph_device_valid)
```

glocaltokens/client.py
```diff
@@ -267,14 +269,16 @@ def find_device(name) -> Optional[GoogleDevice]:
...
-                        google_device=google_device,
-                        hardware=item.hardware.model,
-                    )
+                device = Device(
+                    device_name=item.device_name,
+                    local_auth_token=item.local_auth_token,
+                    google_device=google_device,
+                    hardware=item.hardware.model,
                 )
+                if not device.local_auth_token:
```
leikoilja, line 278: Sorry, I know it's just me being picky, but i think this logic would make more sense if we reverse it:
```python
if device.local_auth_token:
  devices.append(device)   
else:                 
  LOGGER.warning("Device initialization failed, skipping.")
```
and the reason for that is that there might be more conditions when we need to ignore adding a device to a list, thou we shouldn't really care about them all at this point. All we want is "if device is valid - add it to a list, if not - just skip it" :) I hope i make sense

leikoilja (changes requested):
Really nice! Thanks for a lot of very good changes! 
Thou, we are still missing one last test to make sure we test this part:
```python
                if not device.local_auth_token:
                    LOGGER.warning("Device initialization failed, skipping.")
                else:
                    devices.append(device)
```
To make sure if the token is invalid we don't return it in the list :)

tests/test_client.py
ArnyminerZ: I've corrected it a little bit, but the last comparision cannot be done like this:
```python3
self.assertEqual(google_devices[0], homegraph_device_valid)
```
This is because `google_devices` contains `Device`s, and `homegraph_device_valid` is a `Struct`.
We can do this like:
```python3
# Get Homegraph from client
homegraph = self.client.get_homegraph()
homegraph_devices = homegraph.home.devices
self.assertEqual(homegraph_devices[0], homegraph_device_valid)
```
But it would be quite useless because we are overriding that `homegraph.home.devices` in:
```python3
m_get_homegraph.return_value.home.devices = [homegraph_device_valid, homegraph_device_invalid]
```

glocaltokens/client.py
ArnyminerZ: Yeah, totally makes sense, also it's easier to read, so modified it.

tests/test_client.py
leikoilja: ou, i see the issue, thanks for spotting it. 
Yeh, we can totally just compare the local tokens, but i propose we do an extra step and also validate the rest of properties of the device. We can even create a custom assertion for that. @ArnyminerZ, please see the last two commits that i have pushed here as a proposal of a custom assertion class that can help us `assertDevice`. What do u think about it?

leikoilja (approved):
Looks good to me 💯 
What do you think about it, @ArnyminerZ?

ArnyminerZ:
Looks great to me 💥 Merging

PR #48, opened 2021-03-21
Added more debugging for bug-fixing

### LINKED ISSUE #47: Add more logging
There should be a pretty exhaustive debug log in the library, just in case something fails, it's easier to find where it's broken, and find a solution.

ArnyminerZ:
Maybe too redundant, but added everything just in case it is useful. Take a look at it.

Closes #47.

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -345,7 +345,9 @@ class GLocalAuthenticationTokens:
                 else:
                     LOGGER.warning("Device initialization failed, skipping.")
             else:
-                LOGGER.debug("local_auth_token is not initialized")
+                LOGGER.debug(
+                    f"local_auth_token is not initialized for {item.device_name}"
+                )
 
         LOGGER.debug("Google Home devices: {}".format(devices))
 
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -26,6 +26,7 @@ from .google.internal.home.foyer import v1_pb2, v1_pb2_grpc
 from .scanner import GoogleDevice, discover_devices
 from .utils import network as net_utils
 from .utils import token as token_utils
+from .utils.logs import censure
 
 DEBUG = False
 
@@ -47,21 +48,32 @@ class Device:
         """
         Initializes a Device. Can set or google_device or ip and port
         """
+        LOGGER.debug("Initializing new Device instance.")
+        LOGGER.debug(f'Setting self device_name to "{device_name}"')
         self.device_name = device_name
+        LOGGER.debug(f"Setting self local_auth_token to None")
         self.local_auth_token = None
+        LOGGER.debug(f"Setting self google_device to {google_device}")
         self.google_device = google_device
+        LOGGER.debug(f"Setting self hardware to {hardware}")
         self.hardware = hardware
 
         if google_device:
+            LOGGER.debug(f"google_device is not None")
+            LOGGER.debug(f"Setting self ip to {google_device.ip}")
             self.ip = google_device.ip
+            LOGGER.debug(f"Setting self port to {google_device.port}")
             self.port = google_device.port
         else:
+            LOGGER.debug(f"google_device is None")
             if (ip and not port) or (not ip and port):
                 LOGGER.error(
                     "Both ip and port must be set, if one of them is specified."
                 )
                 return
+            LOGGER.debug(f"Setting self ip to {ip}")
             self.ip = ip
+            LOGGER.debug(f"Setting self port to {port}")
             self.port = port
 
         if not local_auth_token:
@@ -88,6 +100,7 @@ class Device:
             LOGGER.error("port must be a valid port")
             return
 
+        LOGGER.debug(f"Setting self local_auth_token to {censure(local_auth_token)}")
         self.local_auth_token = local_auth_token
 
     def __str__(self) -> str:
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -136,9 +136,14 @@ class GLocalAuthenticationTokens:
                 if not set;
 
         """
+        LOGGER.debug("Initializing new GLocalAuthenticationTokens instance.")
+        LOGGER.debug(f'Setting self username to "{username}"')
         self.username: Optional[str] = username
+        LOGGER.debug(f'Setting self password to "{censure(password)}"')
         self.password: Optional[str] = password
+        LOGGER.debug(f'Setting self master_token to "{censure(master_token)}"')
         self.master_token: Optional[str] = master_token
+        LOGGER.debug(f'Setting self android_id to "{censure(android_id)}"')
         self.android_id: Optional[str] = android_id
         if (not self.username or not self.password) and not self.master_token:
             LOGGER.error(
@@ -149,21 +154,28 @@ class GLocalAuthenticationTokens:
         if self.master_token and not token_utils.is_aas_et(self.master_token):
             LOGGER.error("master_token doesn't follow the AAS_ET format")
             return
+        LOGGER.debug(f"Setting self access_token to None")
         self.access_token: Optional[str] = None
+        LOGGER.debug(f"Setting self homegraph to None")
         self.homegraph = None
+        LOGGER.debug(f"Setting self access_token_date to None")
         self.access_token_date: Optional[datetime] = None
+        LOGGER.debug(f"Setting self homegraph_date to None")
         self.homegraph_date: Optional[datetime] = None
 
     @staticmethod
     def _generate_mac_string():
         """Generate random 14 char long string"""
+        LOGGER.debug(f"Generating mac...")
         random_uuid = uuid4()
         random_string = str(random_uuid).replace("-", "")[:ANDROID_ID_LENGTH]
         mac_string = random_string.upper()
+        LOGGER.debug(f"Generated mac: {mac_string}")
         return mac_string
 
     def get_android_id(self) -> Optional[str]:
         if not self.android_id:
+            LOGGER.debug("There is not any stored android_id, generating a new one")
             self.android_id = self._generate_mac_string()
         return self.android_id
 
@@ -177,11 +189,13 @@ class GLocalAuthenticationTokens:
         Get google master token from username and password
         """
         if not self.master_token:
+            LOGGER.debug("There is not any stored master_token, logging in...")
             res = perform_master_login(
                 self.username, self.password, self.get_android_id()
             )
             if "Token" not in res:
                 LOGGER.error("[!] Could not get master token.")
+                LOGGER.debug(f"Request response: {res}")
                 return
             self.master_token = res["Token"]
         LOGGER.debug("Master token: {}".format(self.master_token))
@@ -191,6 +205,9 @@ class GLocalAuthenticationTokens:
         if self.access_token is None or self._has_expired(
             self.access_token_date, ACCESS_TOKEN_DURATION
         ):
+            LOGGER.debug(
+                "There is not any stored access_token, or the stored one has expired, getting a new one..."
+            )
             res = perform_oauth(
                 self.username,
                 self.get_master_token(),
@@ -201,10 +218,12 @@ class GLocalAuthenticationTokens:
             )
             if "Auth" not in res:
                 LOGGER.error("[!] Could not get access token.")
+                LOGGER.debug(f"Request response: {res}")
                 return
             self.access_token = res["Auth"]
             self.access_token_date = datetime.now()
         LOGGER.debug("Access token: {}".format(self.access_token))
+        LOGGER.debug(f"Access token date: {self.access_token_date}")
         return self.access_token
 
     def get_homegraph(self):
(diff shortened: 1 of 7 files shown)


glocaltokens/client.py
```diff
@@ -259,13 +309,20 @@ def find_device(name) -> Optional[GoogleDevice]:
...
         for item in homegraph.home.devices:
             if item.local_auth_token != "":
                 # This checks if the current item is a valid model, only if there are models in models_list.
                 # If models_list is empty, the check should be omitted, and accept all items.
                 if models_list and item.hardware.model not in models_list:
+                    LOGGER.debug("{} not in models_list".format(item.hardware.model))
                     continue
 
+                LOGGER.debug(
+                    "Finding device in network? {}".format(not not network_devices)
```
DurgNomis-drol, line 324: `not not`?

DurgNomis-drol (approved):

glocaltokens/client.py
ArnyminerZ: Yeah, just a shitty way of saying "not None"

glocaltokens/utils/logs.py
```diff
@@ -0,0 +1,8 @@
+def censure(text: str, replace: str = "*") -> str:
```
KapJI, line 1: maybe `censor`?

glocaltokens/scanner.py
```diff
@@ -105,30 +114,50 @@ def discover_devices(
...
+                name,
+                ip,
+                access_port,
+                model,
+            )
             devices.append(GoogleDevice(name, ip, int(access_port), model))
+        else:
+            _LOGGER.debug(
+                'Won\'t add device since model "{}" is not in models_list'.format(model)
+            )
```
KapJI, line 162: @leikoilja is `model_list` still needed?

glocaltokens/client.py
KapJI: that's weird. Why not just `is not None`?

glocaltokens/scanner.py
```diff
@@ -88,15 +89,23 @@ def __init__(self, name: str, ip: str, port: int, model: str):
             _LOGGER.error("port must be an integer value")
             return
 
+        _LOGGER.debug("Setting self name to {}".format(name))
         self.name = name
+        _LOGGER.debug("Setting self ip to {}".format(ip))
         self.ip = ip
+        _LOGGER.debug("Setting self port to {}".format(port))
         self.port = port
+        _LOGGER.debug("Setting self model to {}".format(model))
```
KapJI, line 98: Can this be merged into single message?

glocaltokens/client.py
```diff
@@ -47,21 +48,32 @@ def __init__(
...
         Initializes a Device. Can set or google_device or ip and port
         """
+        LOGGER.debug("Initializing new Device instance.")
+        LOGGER.debug(f'Setting self device_name to "{device_name}"')
         self.device_name = device_name
+        LOGGER.debug(f"Setting self local_auth_token to None")
         self.local_auth_token = None
+        LOGGER.debug(f"Setting self google_device to {google_device}")
         self.google_device = google_device
+        LOGGER.debug(f"Setting self hardware to {hardware}")
```
KapJI, line 58: Can this be merged?

KapJI (changes requested):

glocaltokens/client.py
leikoilja: @KapJI is right, maybe we can log just once at the end somewhat like:
`Initialized new Google device instance with device_name={..}, local_auth_token={..} ....`

glocaltokens/client.py
```diff
@@ -314,10 +374,12 @@ def get_google_devices_json(
 
     def invalidate_access_token(self):
         """Invalidates the current access token"""
+        LOGGER.debug("Invalidating access_token")
```
leikoilja, line 377: Can we move the log after the assignment, that way once we see the log entry we will be sure it's assigned `None` to variables? I know it should not break anywhere here either way, but it kind of makes sense to "do action first, log second"

glocaltokens/scanner.py
leikoilja: Good question, @KapJI.
We have dropped to maintain this list in `google-home`, do you think we still need it in `glocaltokens`, @ArnyminerZ?

glocaltokens/utils/logs.py
```diff
@@ -0,0 +1,8 @@
...
+    """
+    Replaces all the characters in a str by the specified [replace]
+
+    text: The text to censure.
+    replace: The character to instead of the content.
+    """
+    return replace * len(text)
```
leikoilja, line 8: Since we are using this util to print tokens and passwords, we really should add a test to make sure we censor correctly and are not printing any user sensitive data to the log without their consent

leikoilja (changes requested):
Reeeally good work, @ArnyminerZ, thanks a lot! 💣 
We just need to clean it up a little and it's ready to be used!

Just a little question to both @ArnyminerZ and @KapJI, now this package checks global `DEBUG` variable flag and only if it's true it will show `logging.debug`. Is there any near way we can tag along the parent package/component logging level to use here? 
Maybe we can pass an extra optional parameter when we initialize client, somewhat like `verbose` that will set logging level to `debug`? So then on a parent package/integration (google_home in our example) we can pass `verbose` based if a user have `debug` flag set in HA `yaml` configuration for `google_home` component?

KapJI:
yeah, passing an extra flag sounds like a way to go. In HA component we can pass current verbosity level.

glocaltokens/scanner.py
ArnyminerZ: Well, I don't know, it might be useful for someone, maybe. If it doesn't bother, I wouldn't remove it, just in case.

glocaltokens/client.py
```diff
@@ -259,13 +314,20 @@ def find_device(name) -> Optional[GoogleDevice]:
...
         for item in homegraph.home.devices:
             if item.local_auth_token != "":
                 # This checks if the current item is a valid model, only if there are models in models_list.
                 # If models_list is empty, the check should be omitted, and accept all items.
                 if models_list and item.hardware.model not in models_list:
+                    LOGGER.debug("{} not in models_list".format(item.hardware.model))
                     continue
 
+                LOGGER.debug(
+                    "Finding device in network? {}".format(network_devices is not None)
```
leikoilja, line 329: Maybe
"Looking for a device in local network?"

leikoilja, line 329: Maybe "Looking for a device in local network?"

glocaltokens/client.py
```diff
@@ -276,9 +338,12 @@ def find_device(name) -> Optional[GoogleDevice]:
                     hardware=item.hardware.model,
                 )
                 if device.local_auth_token:
+                    LOGGER.debug("Appending device")
```
leikoilja, line 341: Let's be a bit more explicit here 
`LOGGER.debug(f"Appending {device.device_name} device to device list")`

glocaltokens/client.py
```diff
@@ -276,9 +338,12 @@ def find_device(name) -> Optional[GoogleDevice]:
...
                     devices.append(device)
                 else:
                     LOGGER.warning("Device initialization failed, skipping.")
+            else:
+                LOGGER.debug("local_auth_token is not initialized")
```
leikoilja, line 346: Let's be a bit more explicit here
`LOGGER.debug(f"local_auth_token is not initialized for {device.device_name}")`

glocaltokens/scanner.py
```diff
@@ -92,11 +93,20 @@ def __init__(self, name: str, ip: str, port: int, model: str):
         self.ip = ip
         self.port = port
         self.model = model
+        _LOGGER.debug(
+            "Set self name to {}, ip to {}, port to {} and model to {}".format(
+                name, ip, port, model
+            )
```
leikoilja, line 99: Let's make use of f-strings here as well instead of `format`, so we have it consistent everywhere :)

glocaltokens/scanner.py
```diff
@@ -92,11 +93,20 @@ def __init__(self, name: str, ip: str, port: int, model: str):
...
+        )
 
         if not 0 <= self.port <= 65535:
             _LOGGER.error("Port is out of the valid range: [0,65535]")
             return
 
+    def __str__(self) -> str:
+        """Serializes the class into a str"""
+        return f"{{name:{self.name},ip:{self.ip},port:{self.port},model:{self.model}}}"
```
leikoilja, line 108: Is this actually working?

glocaltokens/scanner.py
leikoilja: Since we fallback to fetching all devices if no `model_list` is provided we can just keep it i guess :)

tests/test_utils.py
```diff
@@ -0,0 +1,21 @@
+from unittest import TestCase
+
+from faker import Faker
+
+from glocaltokens.utils.logs import censor
+
+faker = Faker()
+
+
+class GLocalAuthenticationTokensClientTests(TestCase):
```
leikoilja, line 10: This is not a client test. Please name it something like `UtilsTests`

tests/test_utils.py
```diff
@@ -0,0 +1,21 @@
...
+    def setUp(self):
+        """Setup method run before every test"""
+        pass
+
+    def tearDown(self):
+        """Teardown method run after every test"""
+        pass
+
+    def test_logs(self):
+        random = faker.word()
```
leikoilja, line 20: Shall we name it `secret_string` or smth?

leikoilja (changes requested):
In general looks good to me. Please fix the failing tests and see some minor comments.

@KapJI do you want to take a look as well? :)

glocaltokens/client.py
KapJI: Or `Adding {device.device_name} to devices list`.

KapJI:
Looks good to me overall

glocaltokens/scanner.py
ArnyminerZ: Haven't tested, but it should, is there any visible issue? I've just added a test and it does

tests/test_utils.py
ArnyminerZ: Oops, just a copy-paste typo there.

leikoilja (changes requested):
Looks great to me! 
@ArnyminerZ can you please fix failing tests that are broken now after you have added `censor` method? :) 

Also, do you mind us finding a way to pass the `verbose` parameter her to the `GLocalAuthenticationTokens` that will set the logger level to `debug`? I can do that tomorrow if you'd like :)

leikoilja:
@ArnyminerZ, @KapJI just pushed test fixes and `verbose` logging argument. Please take a look if you guys agree with it :)

glocaltokens/client.py
```diff
@@ -123,6 +120,7 @@ def __init__(
         password: Optional[str] = None,
         master_token: Optional[str] = None,
         android_id: Optional[str] = None,
+        verbose: Optional[bool] = False,
```
KapJI, line 123: It's not `Optional`, just `bool`. `Optional` is `None` or `type`.

glocaltokens/utils/logs.py
```diff
@@ -1,8 +1,10 @@
-def censor(text: str, replace: str = "*") -> str:
+def censor(text: str, char: str = "*") -> str:
     """
     Replaces all the characters in a str by the specified [replace]
 
     text: The text to censure.
     replace: The character to instead of the content.
```
KapJI: Update this.

glocaltokens/utils/logs.py
```diff
@@ -1,8 +1,10 @@
...
     """
-    return replace * len(text)
+    text = text if text else ""
```
KapJI, line 8: Why do you need this? `text` is specified as `str`, not `Optional[str]`.

glocaltokens/utils/logs.py
```diff
@@ -1,8 +1,10 @@
-def censor(text: str, replace: str = "*") -> str:
+def censor(text: str, char: str = "*") -> str:
```
KapJI, line 1: How about removing `char` argument completely? Why is it needed?

glocaltokens/utils/logs.py
```diff
@@ -1,8 +1,10 @@
...
     """
     Replaces all the characters in a str by the specified [replace]
 
     text: The text to censure.
     replace: The character to instead of the content.
     """
-    return replace * len(text)
+    text = text if text else ""
+    censored_text = text[0] + (len(text) - 1) * char if len(text) else text
+    return censored_text
```
KapJI, line 10:
```python3
return text[0] + (len(text) - 1) * char if text else text
```

KapJI (changes requested):

glocaltokens/utils/logs.py
leikoilja: We have a problem here that it can we `None` when we call it from client passing in `master_token` for example

glocaltokens/utils/logs.py
leikoilja: Good point, we don't really need it to be customizable :D

glocaltokens/utils/logs.py
leikoilja: haha you really don't like my "1. assignment; 2. return" :D will fix it thou ;)

glocaltokens/utils/logs.py
KapJI: Then make text `Optional[str]` 🙂

KapJI (approved):
Thanks! 🚀

PR #49, opened 2021-03-24
Setting up poetry

leikoilja:
- Setting up poetry
- Updating github actions to use poetry
- Updating contributions file
- Running pre-commit linters on all files


KapJI:
I'm not sure if `tox` is still useful with poetry. Testing with different python version can be done using Github Actions and I think it's better just to replace it `pytest`.

pyproject.toml
```diff
@@ -0,0 +1,41 @@
+[tool.poetry]
+name = "glocaltokens"
+version = "0.0.0"
```
KapJI, line 3: This should be an actual package version. Basically poetry should replace setuptools and `setup.py`.

KapJI: This can be done in a separate PR to split this migration in parts.

leikoilja: oh that's nice!

leikoilja: @KapJI I have just tried `poetry build` locally and it works flawlessly, super nice!

leikoilja:
@KapJI, what you propose with `tox` sounds reasonable. But how would that look practically?
- Removing `tox` and `tox-gh-actions` from poetry
- Setting up github action to run tox independently from poetry?

KapJI:
I mean removing `tox` at all and run `pytest` directly from `poetry` 🙂

pyproject.toml
KapJI: Then you should be also able to use `poetry publish` to release the new version.

leikoilja: Exactly! Thanks for introducing `poetry`, such a time saver :P

leikoilja:
Roger that. FIxed it and i like it now :) 
What do you think, @KapJI?
As next PR we can enable mypy on this repo i guess :)

.github/workflows/tests.yaml
```diff
@@ -28,6 +28,6 @@ jobs:
       - name: Install dependencies
         run: poetry install
 
-      - name: Run tox
+      - name: Run pytest
         run: |
-          poetry run tox
+          poetry run pytest
```
KapJI: Are're already running tests in linting job, this one can be removed.
You can rename linting to testing as well.

.pre-commit-config.yaml
```diff
@@ -16,3 +16,11 @@ repos:
         args:
           - --safe
           - --quiet
```
KapJI, line 18: Replace `black` and `isort` entries with ones which run versions installed by poetry. Similar to `ha-google-home`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,89 @@
...
+
+Activate `pre-commit` git hook:
+
+```console
+$ poetry run pre-commit install
+```
+
+Now the pre-commit tests will be done every time you commit.
+
+You can run the tests on all repository file with the command:
```
KapJI, line 81:
```suggestion
You can also run the tests on all repository files manually with the command:
```

pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+pre-commit = "^2.11.1"
+pylint = "^2.7.2"
+isort = "^5.8.0"
+mypy = "^0.812"
+pytest = "^6.2.2"
+Faker = "^6.6.2"
+mock = "^4.0.3"
+
+[tool.poetry.urls]
+"Bug Tracker" = "https://github.com/leikoilja/glocaltokens/issues"
```
KapJI, line 31: Delete `HISTORY.md` and add link to release notes here.

pyproject.toml
```diff
@@ -0,0 +1,43 @@
+[tool.poetry]
+name = "glocaltokens"
+version = "0.2.6"
+description = "Tool to extract Google device local authentication tokens in Python"
+authors = ["Ilja Leiko <leikoilja@gmail.com>"]
+license = "MIT"
+repository = "https://github.com/leikoilja/glocaltokens"
```
KapJI, line 7: Not sure if you need it but you can also add a `homepage` here.

pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+readme = "README.md"
```
KapJI, line 8: You can add `keywords` in this section to match old `setup.py`. Although I'm not sure how useful they are.

KapJI: Also it's worth adding `classifiers` here. See [poetry documentation](https://python-poetry.org/docs/pyproject/) and [packaging documentation](https://packaging.python.org/guides/distributing-packages-using-setuptools/#classifiers).

KapJI (changes requested):
Few final touches to fully deprecate `setup.py`.

pyproject.toml
```diff
@@ -1,11 +1,24 @@
...
+keywords = [
+    "Authentication",
+    "Google Home",
+    "Google",
+    "LocalAuthenticationTokens",
+]
 license = "MIT"
-repository = "https://github.com/leikoilja/glocaltokens"
+name = "glocaltokens"
 readme = "README.md"
```
KapJI: Is it reordered by you or some tool? I think previous ordering was better with name and version on the top and links next to each other.

KapJI (approved):
Other than that, looks great!

tox.ini
```diff
@@ -1,15 +1,15 @@
 [tox]
```
KapJI, line 1: Missed that. `tox.ini` should be removed as tox is no longer used.

pyproject.toml
leikoilja: Just reordered it manually to be alphabetic :)

KapJI: I think this section is better to be semantically sorted 🙂

leikoilja:
Awesome, thanks, @KapJI for help! 💣 Let's merge it in! 💥

Issue #46, opened 2021-03-21
`Castlistner` should implement `remove_service`


DurgNomis-drol:
**Describe the bug**

```text
2021-03-21 10:56:13 ERROR (zeroconf-ServiceBrowser__googlecast._tcp.local._264) [root] Uncaught thread exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.8/site-packages/zeroconf/__init__.py", line 1760, in run
    self._service_state_changed.fire(
  File "/usr/local/lib/python3.8/site-packages/zeroconf/__init__.py", line 1513, in fire
    h(**kwargs)
  File "/usr/local/lib/python3.8/site-packages/zeroconf/__init__.py", line 1618, in on_change
    listener.remove_service(*args)
AttributeError: 'CastListener' object has no attribute 'remove_service'
```

See this link [Homeassistant/core#48117](https://github.com/home-assistant/core/issues/48117)

**To Reproduce**

Appears first time i restart, but only once.

**Expected behavior**

No error message.

**Screenshots**

**Additional context**

leikoilja:
Really good find, thank you, @DurgNomis-drol 💥

KapJI:
I'd say it's more a bugfix rather that feature :)

leikoilja:
Uuu, you are right, @KapJI, thanks :)

PR #50, opened 2021-03-25
Setting up mypy

### LINKED ISSUE #43: Setting up poetry dependancy management and packaging tool
We need to setup `poetry` to manage the package.

- [ ] set up pre commit hooks for `mypy`
- [ ] set up CI job for `mypy`

leikoilja:
The beginning of setting up mypy

- sets up pre commit hooks for mypy
- sets up CI job for mypy

Closes #43

### DIFF
--- a/.github/workflows/linting-and-testing.yaml
+++ b/.github/workflows/linting-and-testing.yaml
@@ -33,7 +33,3 @@ jobs:
       - name: Run pre-commit on all files
         run: |
           poetry run pre-commit run --all-files --show-diff-on-failure --color=always
-
-      # - name: Run MyPy
-      #   run: |
-      #     poetry run mypy glocaltokens
--- a/.github/workflows/linting-and-testing.yaml
+++ b/.github/workflows/linting-and-testing.yaml
@@ -33,3 +33,7 @@ jobs:
       - name: Run pre-commit on all files
         run: |
           poetry run pre-commit run --all-files --show-diff-on-failure --color=always
+
+      # - name: Run MyPy
+      #   run: |
+      #     poetry run mypy glocaltokens
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -206,7 +206,7 @@ class GLocalAuthenticationTokens:
             if "Token" not in res:
                 LOGGER.error("[!] Could not get master token.")
                 LOGGER.debug(f"Request response: {res}")
-                return ""
+                return None
             self.master_token = res["Token"]
         LOGGER.debug("Master token: {}".format(self.master_token))
         return self.master_token
@@ -230,7 +230,7 @@ class GLocalAuthenticationTokens:
             if "Auth" not in res:
                 LOGGER.error("[!] Could not get access token.")
                 LOGGER.debug(f"Request response: {res}")
-                return ""
+                return None
             self.access_token = res["Auth"]
             self.access_token_date = datetime.now()
         LOGGER.debug(
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -47,8 +47,11 @@ class Device:
         LOGGER.debug("Initializing new Device instance.")
         self.device_name = device_name
         self.local_auth_token = None
+        self.ip = ip
+        self.port = port
         self.google_device = google_device
         self.hardware = hardware
+
         LOGGER.debug(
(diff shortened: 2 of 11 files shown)


leikoilja:
@KapJI, it's a bit over my head, can you please help me to understand why `exclude = 'glocaltokens/google/*'` line in mypy config is ignored? 🤔  This results in checking proto files that ends up with tons of errors

KapJI:
Add `__init__.py` in all `google` subdirectories. And then in `mypy.ini` (I actually suggest to put configuration if `setup.cfg`) add:
```
[mypy-glocaltokens.google.*]
ignore_errors = True
```

.github/workflows/linting-and-testing.yaml
```diff
@@ -33,3 +33,7 @@ jobs:
       - name: Run pre-commit on all files
         run: |
           poetry run pre-commit run --all-files --show-diff-on-failure --color=always
+
+      # - name: Run MyPy
+      #   run: |
+      #     poetry run mypy glocaltokens
```
KapJI, line 39: With `mypy` in `pre-commit` config, you won't need this.

leikoilja:
@KapJI, i m working through fixing `mypy` errors and there is one i don't know how to fix. Can you please help me with that one:
```bash
glocaltokens/scanner.py:135: error: Argument 3 to "ServiceBrowser" has incompatible type "CastListener"; expected "Union[ServiceListener, List[Callable[..., None]]
, None]"
Found 1 error in 1 file (checked 4 source files)
```

KapJI:
Derive `CastListener` from [`zeroconf.ServiceListener`](https://github.com/jstasiak/python-zeroconf/blob/4da1612b728acbcf2ab0c4bee09891c46f387bfb/zeroconf/__init__.py#L1531)

KapJI:
It's actually shows that `CastListener` needs to implement `remove_service`. We had issue for that but I can't find it.

DurgNomis-drol:
#46

KapJI:
Yep, this one :)

KapJI:
Also take a look at [`@overrides` decorator](https://github.com/mkorpela/overrides) to check that you're actually overriding that method.

leikoilja:
Mypy doesn't argue on not having `remove_service`, therefore I'll keep it for the next PR

glocaltokens/client.py
```diff
@@ -200,7 +203,7 @@ def get_master_token(self) -> Optional[str]:
             if "Token" not in res:
                 LOGGER.error("[!] Could not get master token.")
                 LOGGER.debug(f"Request response: {res}")
-                return
+                return ""
```
KapJI, line 206:
```suggestion
                return None
```

glocaltokens/client.py
```diff
@@ -223,7 +226,7 @@ def get_access_token(self) -> Optional[str]:
             if "Auth" not in res:
                 LOGGER.error("[!] Could not get access token.")
                 LOGGER.debug(f"Request response: {res}")
-                return
+                return ""
```
KapJI, line 229:
```suggestion
                return None
```

glocaltokens/client.py
```diff
@@ -313,7 +316,7 @@ def find_device(name) -> Optional[GoogleDevice]:
                     return device
             return None
 
-        devices: [Device] = []
+        devices: List[Device] = []
```
KapJI, line 319: Good catch 😀

mypy.ini
```diff
@@ -0,0 +1,5 @@
+[mypy]
+ignore_missing_imports = True
```
KapJI, line 2: Let's only set this for packages which doesn't provide typing info, e.g. `faker`. Like in `ha-google-home`.

KapJI: Worth setting `python_version = 3.8` here.

mypy.ini
```diff
@@ -0,0 +1,5 @@
+[mypy]
```
KapJI, line 1: Let's move this to `setup.cfg`. The idea is to have less configuration files if we add more tools. All python tools support configuration either in `setup.cfg` or `pyproject.toml` (which should be preferred but not supported by all tools yet).

leikoilja:
Check it out, @KapJI

KapJI (approved):
Awesome! Ship it 🚀

PR #56, opened 2021-04-05
Enabling pylint

leikoilja:
Enabling pylint and linting

Breaking change: changing `device.ip` to `device.ip_address`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
       - id: pylint
         name: pylint
         entry: poetry run pylint
-        exclude: ^glocaltokens/
+        exclude: ^glocaltokens/google/
         language: system
         types: [python]
       - id: pytest
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,11 +31,12 @@ repos:
         exclude: glocaltokens/google/
         language: system
         types: [python]
-      # - id: pylint
-      #   name: pylint
-      #   entry: poetry run pylint
-      #   language: system
-      #   types: [python]
+      - id: pylint
+        name: pylint
+        entry: poetry run pylint
+        exclude: ^glocaltokens/
+        language: system
+        types: [python]
       - id: pytest
         name: pytest
         language: system
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -33,7 +33,6 @@ logging.basicConfig(level=logging.ERROR)
 LOGGER = logging.getLogger(__name__)
 
 
-# pylint: disable=too-many-arguments
 class Device:
     """Device representation"""
 
@@ -126,7 +125,6 @@ class Device:
         }
 
 
-# pylint: disable=too-many-instance-attributes
 class GLocalAuthenticationTokens:
     """Client"""
 
@@ -293,18 +291,16 @@ class GLocalAuthenticationTokens:
                 self.homegraph_date = datetime.now()
             except grpc.RpcError as rpc_error:
                 LOGGER.debug("Got an RpcError.")
-                # pylint: disable=no-else-return
                 # pylint: disable=no-member
                 if rpc_error.code().name == "UNAUTHENTICATED":
                     LOGGER.warning("The access token has expired. Getting a new one.")
                     self.invalidate_access_token()
                     return self.get_homegraph()
-                else:
-                    LOGGER.error(
-                        "Received unknown RPC error: code=%s " "message=%s",
-                        rpc_error.code(),
-                        rpc_error.details(),
-                    )
+                LOGGER.error(
+                    "Received unknown RPC error: code=%s " "message=%s",
+                    rpc_error.code(),
+                    rpc_error.details(),
+                )
         return self.homegraph
 
     def get_google_devices(
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -1,3 +1,5 @@
+"""Client"""
+
 from datetime import datetime
 import json
 import logging
@@ -31,13 +33,16 @@ logging.basicConfig(level=logging.ERROR)
 LOGGER = logging.getLogger(__name__)
 
 
+# pylint: disable=too-many-arguments
 class Device:
+    """Device representation"""
+
     def __init__(
         self,
         device_name: str,
         local_auth_token: str,
         google_device: Optional[GoogleDevice] = None,
-        ip: Optional[str] = None,
+        ip_address: Optional[str] = None,
         port: Optional[int] = None,
         hardware: Optional[str] = None,
     ):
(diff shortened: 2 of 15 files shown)


.pre-commit-config.yaml
```diff
@@ -31,11 +31,12 @@ repos:
...
         types: [python]
-      # - id: pylint
-      #   name: pylint
-      #   entry: poetry run pylint
-      #   language: system
-      #   types: [python]
+      - id: pylint
+        name: pylint
+        entry: poetry run pylint
+        exclude: ^glocaltokens/
```
KapJI, line 37: Let's enable it for the whole project, but disable checks globally in `pyproject.toml` which are failing right now. This way `glocaltokens` will have at least some coverage.

tests/assertions.py
```diff
@@ -1,12 +1,21 @@
+"""
+Common assertion helper classes used for unittesting
+"""
+# pylint: disable=E1101
```
KapJI, line 4: Please use human readable names.
```suggestion
# pylint: disable=no-member
```

KapJI (changes requested):
Add black compatible configuration in `pyproject.toml` (`[tool.pylint.format]`). You can copy it from `ha-google-home`.

.pre-commit-config.yaml
leikoilja: That reg-ex was meant to me `^glocaltokens/google/` to avoid checking proto files. I would really want to put `exclude` in the `pyproject.toml` file, but when i put it into `pyptoject.toml` it `exclude` doesn't work
```
[tool.pylint.format]
max-line-length = 88
exclude = "^glocaltokens/google/"
```

KapJI: First of all it should be in [tools.pylint.master] section. Also options are called ignore (or ignore-patterns).

Try:

[tools.pylint.master]
ignore=google

leikoilja: Having a wrong regex also well explains why i didn't need to change anything around the main logic in the client and scanner, that got me confused at first

glocaltokens/client.py
```diff
@@ -47,34 +52,35 @@ def __init__(
...
-            self.ip = google_device.ip
-            LOGGER.debug(f"Setting self port to {google_device.port}")
+            LOGGER.debug("Setting self ip to %s", google_device.ip_address)
+            self.ip_address = google_device.ip_address
+            LOGGER.debug("Setting self port to %s", google_device.port)
             self.port = google_device.port
         else:
             LOGGER.debug("google_device is None")
-            if (ip and not port) or (not ip and port):
+            if (ip_address and not port) or (not ip_address and port):
```
KapJI, line 76: Let's cover all cases.
```suggestion
            if not ip_address or not port:
```

.pre-commit-config.yaml
```diff
@@ -31,11 +31,12 @@ repos:
...
         types: [python]
-      # - id: pylint
-      #   name: pylint
-      #   entry: poetry run pylint
-      #   language: system
-      #   types: [python]
+      - id: pylint
+        name: pylint
+        entry: poetry run pylint
+        exclude: ^glocaltokens/google/
```
KapJI, line 37: I still think it's better to move this to `pyproject.toml`. This way you can run `pylint` without `pre-commit`.

glocaltokens/client.py
```diff
@@ -180,10 +192,11 @@ def _generate_mac_string():
         random_uuid = uuid4()
         random_string = str(random_uuid).replace("-", "")[:ANDROID_ID_LENGTH]
         mac_string = random_string.upper()
-        LOGGER.debug(f"Generated mac: {mac_string}")
+        LOGGER.debug("Generated mac: %s", mac_string)
```
KapJI, line 195: Btw, I found this code is generating incorrect mac address. It should be 12 chars long, not 14. And the way you geneate it is... creative.

glocaltokens/client.py
```diff
@@ -272,14 +282,17 @@ def get_homegraph(self):
                 self.homegraph_date = datetime.now()
             except grpc.RpcError as rpc_error:
                 LOGGER.debug("Got an RpcError.")
+                # pylint: disable=no-else-return
```
KapJI, line 285: Why not just dropping the `else`?

glocaltokens/scanner.py
```diff
@@ -1,3 +1,6 @@
+"""Zeroconf based scanner"""
+# pylint: disable=arguments-differ
```
KapJI, line 2: This sounds like a bug.

glocaltokens/scanner.py
```diff
@@ -82,25 +85,32 @@ def get_value(key):
             callback()
 
 
+# pylint: disable=too-few-public-methods
```
KapJI, line 88: Disable it globally in `pyproject.toml`.

glocaltokens/scanner.py
```diff
@@ -109,21 +119,27 @@ def __init__(self, name: str, ip: str, port: int, model: str):
...
     def __str__(self) -> str:
         """Serializes the class into a str"""
-        return f"{{name:{self.name},ip:{self.ip},port:{self.port},model:{self.model}}}"
+        return (
+            f"{{name:{self.name},ip:{self.ip_address},"
+            f"port:{self.port},model:{self.model}}}"
+        )
 
 
+# pylint: disable=too-many-locals
```
KapJI, line 128: Move to `pyproject.toml`.

glocaltokens/scanner.py
```diff
@@ -109,21 +119,27 @@ def __init__(self, name: str, ip: str, port: int, model: str):
...
+# pylint: disable=unused-argument
```
KapJI, line 129: Is it a public API? Can you just rename the argument?

glocaltokens/utils/network.py
```diff
@@ -1,9 +1,11 @@
+"""Network utilities"""
 import socket
 
 import glocaltokens.utils.types as type_utils
 
 
 def is_valid_ipv4_address(address):
+    """Check if ip address is ipv4 type"""
```
KapJI, line 8:
```suggestion
    """Check if ip address is ipv4"""
```

glocaltokens/utils/network.py
```diff
@@ -19,6 +21,7 @@ def is_valid_ipv4_address(address):
 
 
 def is_valid_ipv6_address(address):
+    """Check if ip address is ipv6 type"""
```
KapJI, line 24:
```suggestion
    """Check if ip address is ipv6"""
```

tests/test_client.py
```diff
@@ -230,6 +241,7 @@ def test_get_access_token(self, m_perform_oauth, m_get_master_token, m_log):
         access_token = self.client.get_access_token()
         self.assertEqual(m_perform_oauth.call_count, 1)
 
+    # pylint: disable=too-many-arguments
```
KapJI, line 244: Move to `pyproject.toml`.

KapJI (changes requested):

glocaltokens/client.py
leikoilja: Agree that the logic here can be improved, but let's keep it for another PR, cause it will also involve changing tests

.pre-commit-config.yaml
leikoilja: sorry i dont think i follow what you mean. 
If i add
```
[tools.pylint.master]
ignore = "google"
```
to `pyproject.toml` how to i run `pylint`? `poetry run pylint glocaltokens` still includes `google` directory and without `exclude` in `pre-commit-config` precommit `pylint` hook also fails on `google` dir

glocaltokens/client.py
leikoilja: hmmmm, we should look into that, maybe it would even be better to entirely switch to iOS mac which is uuid4 consistent format

glocaltokens/scanner.py
leikoilja: Well, not really, it's just overriding the listener method following the `pycast` example:
```
************* Module glocaltokens.scanner
glocaltokens/scanner.py:33:4: W0221: Parameters differ from overridden 'add_service' method (arguments-differ)
glocaltokens/scanner.py:38:4: W0221: Parameters differ from overridden 'update_service' method (arguments-differ)
glocaltokens/scanner.py:43:4: W0221: Parameters differ from overridden 'remove_service' method (arguments-differ)
```

glocaltokens/client.py
KapJI: How about just generating 6 random bytes?
```python
"".join([f"{random.randrange(256):02x}" for _ in range(6)])
```

glocaltokens/scanner.py
KapJI: Yes, and this is a potential bug if someone decide to use kwargs. You should use the same names for args:
https://github.com/jstasiak/python-zeroconf/blob/master/zeroconf/__init__.py#L1541

.pre-commit-config.yaml
KapJI: Ok, let's add an `exclude` here, I'll take a look how to do this in pylint config later.

glocaltokens/client.py
leikoilja: Sounds good, let's have a separate ticket for that

KapJI (approved):
Thanks! Let's move on with improving logging. That issue with redacting tokens from logs is pretty popular.

leikoilja:
Exactly, really wanna merge it quick so it's in a new release, @KapJI

PR #63, opened 2021-04-06
Update gpsoauth to 1.0.0

KapJI:
It was recently updated to 1.0.0 and this includes few important changes:
- List of used ciphers is now hardcoded, this should make authorisation more reliable.
- It's strictly typed and exports typing information. This highlighted few bugs when username and password were not checked.

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -174,7 +174,7 @@ class GLocalAuthenticationTokens:
         )
 
     @staticmethod
-    def _generate_mac_string():
+    def _generate_mac_string() -> str:
         """Generate random 14 char long string"""
         LOGGER.debug("Generating mac...")
         random_uuid = uuid4()
@@ -183,7 +183,7 @@ class GLocalAuthenticationTokens:
         LOGGER.debug(f"Generated mac: {mac_string}")
         return mac_string
 
-    def get_android_id(self) -> Optional[str]:
+    def get_android_id(self) -> str:
         if not self.android_id:
             LOGGER.debug("There is not any stored android_id, generating a new one")
             self.android_id = self._generate_mac_string()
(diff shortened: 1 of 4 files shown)


glocaltokens/client.py
```diff
@@ -198,6 +198,9 @@ def get_master_token(self) -> Optional[str]:
         """
         Get google master token from username and password
         """
+        if self.username is None or self.password is None:
+            LOGGER.error("Username and password are not set.")
+            return None
```
leikoilja, line 203: We shouldn't get to this point, cause we have checks in `__init__` that prevents creating the client credentials are not supported. Should it not be enough?

glocaltokens/client.py
```diff
@@ -219,9 +222,16 @@ def get_access_token(self) -> Optional[str]:
                 "There is not any stored access_token, "
                 "or the stored one has expired, getting a new one..."
             )
+            master_token = self.get_master_token()
+            if master_token is None:
+                LOGGER.debug("Unable to obtain master token.")
+                return None
+            if self.username is None:
+                LOGGER.error("Username is not set.")
+                return None
```
leikoilja, line 231: We shouldn't get to this point, cause we have checks in `__init__` that prevents creating the client credentials are not supported. Should it not be enough?

pyproject.toml
```diff
@@ -24,7 +24,7 @@ repository = "https://github.com/leikoilja/glocaltokens"
 
 [tool.poetry.dependencies]
 python = "^3.8"
-gpsoauth = "^0.4.3"
+gpsoauth = "^1.0.0"
```
leikoilja, line 27: Awesome work with all you have done for v1 release! 💥  🚀

leikoilja (approved):
LGTM, but not sure we need all the explicit checks for username/password

glocaltokens/client.py
KapJI: It's pretty hard to express in python type system "A or B is set". For "xor" it's possible to use `Union` but "or" is harder. It's better to add explicit checks even if they always will be false. Theoretically it's not forbidden to modify fields after object was created, so it's possible to trigger this even with strict mypy checks.

PR #69, opened 2021-04-11
Reduce discovery timeout

KapJI:
I noticed in my logs and logs of other people that zeroconf discovery is actually pretty fast and all devices are discovered under 1 second.

Reducing this timeout will make `google_home` initial setup and component startup much faster.

Also add an option to `get_google_devices` to override it.

For some reason `pylint` was broken because it was unable to find `socket` symbols which are implemented in C. Fixed that as well.

And `decorator==5.0.1` no longer exists: https://pypi.org/project/decorator/#history

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -15,6 +15,7 @@ from .const import (
     ACCESS_TOKEN_DURATION,
     ACCESS_TOKEN_SERVICE,
     ANDROID_ID_LENGTH,
+    DISCOVERY_TIMEOUT,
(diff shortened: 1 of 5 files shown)


leikoilja (approved):
That's nice! 
The only this is I'm afraid we might start having tons of timeout if ppl network is not tuned well... I think we can totally take that "risk" but let's keep an eye for timeouts in logs, so we accidentally don't introduce a potential problem 😅

leikoilja:
Thank you 💥 🚀

PR #70, opened 2021-04-11
Set Dependabot versioning strategy

KapJI:
As we discussed, set versioning stratagy to lockfile only to prevent changed in `pyproject.toml` and preserve library versioning constraints. This should prevent `grpcio` updates. Also as a library `glocaltokens` should have more relaxed versioning requirements.

Also import updated Labeler config from `ha-google-home`.


leikoilja (approved):
awesome! thanks for doing that 🚀

PR #71, opened 2021-04-11
Add more type hints, fix some pylint warnings

### LINKED ISSUE #66: Move pylint exclude to pyproject
Let's move pylint exclude from precommits to pyproject.toml

See https://github.com/leikoilja/glocaltokens/pull/56#discussion_r608507433

KapJI:
Mypy helps to catch places where `Optional` is not handled properly.

Fixes #66

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -279,7 +279,6 @@ class GLocalAuthenticationTokens:
         )
         return self.access_token
 
-    # pylint: disable=no-member
     def get_homegraph(self):
         """Returns the entire Google Home Foyer V2 service"""
         if self.homegraph is None or self._has_expired(
@@ -322,7 +321,10 @@ class GLocalAuthenticationTokens:
                 self.homegraph_date = datetime.now()
             except grpc.RpcError as rpc_error:
                 LOGGER.debug("%s Got an RpcError", log_prefix)
-                if rpc_error.code().name == "UNAUTHENTICATED":
+                if (
+                    rpc_error.code().name  # pylint: disable=no-member
+                    == "UNAUTHENTICATED"
+                ):
                     LOGGER.warning(
                         "%s The access token has expired. Getting a new one.",
                         log_prefix,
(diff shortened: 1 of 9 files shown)


glocaltokens/client.py
```diff
@@ -289,20 +288,26 @@ def get_homegraph(self):
                 "There is no stored homegraph, or it has expired, getting a new one..."
             )
             log_prefix = "[GRPC]"
+            access_token = self.get_access_token()
+            if not access_token:
+                LOGGER.debug("%s Unable to obtain access token.", log_prefix)
+                return None
```
leikoilja, line 294: I'm wondering if we should extract that check into a helper function decorator or just a helper function to call at all places as we do now

pyproject.toml
```diff
@@ -51,12 +51,21 @@ ipdb = "^0.13.7"
 extension-pkg-whitelist = [
     "_socket",
 ]
+ignore = ["google"]
```
leikoilja, line 54: nice!

pyproject.toml
```diff
@@ -51,12 +51,21 @@ ipdb = "^0.13.7"
...
+
+[tool.pylint.basic]
+good-names = [
+    "zc",
+]
```
leikoilja, line 59: What is that for? 🤔

tests/test_client.py
```diff
@@ -250,7 +251,7 @@ def test_get_access_token(self, m_perform_oauth, m_get_master_token, m_log):
     @patch("glocaltokens.client.GLocalAuthenticationTokens.get_access_token")
     def test_get_homegraph(
         self,
-        m_get_access_token,  # pylint: disable=unused-argument
+        _m_get_access_token,
```
leikoilja, line 254: uuuu, didn't know you can do that 👍

tests/test_client.py
```diff
@@ -279,6 +280,7 @@ def test_get_homegraph(
         self.assertEqual(m_get_home_graph_request.call_count, 1)
 
         # Expired homegraph
+        assert self.client.homegraph_date is not None
```
leikoilja, line 283: Let's have here `unittest` pattern, just to have it consistent with the rest of the code?
```suggestion
        self.assertIsNotNone(self.client.homegraph_date)
```

leikoilja (approved):
Thanks, those are some really good changes 💯

pyproject.toml
KapJI: To fix `invalid-name` pylint warning to `ServiceListener` methods.

tests/test_client.py
KapJI: Unfortunately, `mypy` can't narrow type from such calls: https://github.com/python/mypy/issues/4063

PR #77, opened 2021-04-14
Flake8 plugins, disable Dependabot labels, fix Labeler

KapJI:
Backport some changes from `ha-google-home`:

- Install Flake8 plugins
- Updated Labeler config with proper globs

Apart from that:

- Fix github-actions updates with Dependabot. I think it doesn't work with `"/.github/workflows"`, at least I haven't seen any PRs for github actions. `"/"` should work for sure.
- I found in [Dependabot sources](https://github.com/dependabot/dependabot-core/blob/2f0db3e851ba2cc43d0b6dcd70da5e69d5b63eb6/common/lib/dependabot/pull_request_creator/labeler.rb#L204) how to disable version labels.


.github/labels.yml
```diff
@@ -56,6 +56,11 @@
   description: Performance
   color: "016175"
 
+# Prevent Dependabot from adding version labels
+- name: skip-release
+  description: Skip release
+  color: "000000"
+
```
leikoilja, line 63: Great find! 
According to their source just by having this label will prevent dependabot from increasing the version, right?

leikoilja (approved):
Awesome stuff! Thanks

.github/labels.yml
KapJI: [Yep!](https://www.youtube.com/watch?v=sGXQR5Xbpqs)

leikoilja: haha good one :D

PR #80, opened 2021-04-14
Revert "Cleanup depdendabot config"

KapJI:
Reverts leikoilja/glocaltokens#79

It actually prevents from adding language label 🙁


leikoilja (approved):
what's language label?

KapJI:
Like `github_actions` here: https://github.com/leikoilja/ha-google-home/pull/207#issuecomment-819464821

PR #82, opened 2021-04-14
Generate mypy stubs for protobuf

KapJI:
I used [mypy-protobuf](https://github.com/dropbox/mypy-protobuf) to generate these stubs.

With this we can enable strict checks for the library (but not for tests yet).


leikoilja (approved):
That's super nice! Thanks 🚀

PR #83, opened 2021-04-15
Add type hints to tests, enable strict checks

KapJI:
Enforce the same strict checks as we do in `ha-google-home`.


leikoilja (approved):
Yay for STRICT 🚀  Thanks

PR #84, opened 2021-04-15
Fix MAC address generation

### LINKED ISSUE #65: Improve android id generation
Improve how we generate android id and double check what is the required length.

See https://github.com/leikoilja/glocaltokens/pull/56#discussion_r608499319

KapJI:
Generate 12 random chars.

Closes #65

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -217,7 +217,7 @@ class GLocalAuthenticationTokens:
         LOGGER.debug("Generating android id...")
         mac_string = "".join(
             [f"{random.randrange(16):x}" for _ in range(ANDROID_ID_LENGTH)]
-        ).upper()
+        )
         LOGGER.debug("Generated android id: %s", mac_string)
         return mac_string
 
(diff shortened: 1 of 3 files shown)


leikoilja:
Are we sure it's 12 chars? 🤔

leikoilja:
Just tried on friend's samsung s10+ and serial number is 11 chars

KapJI:
It should be 6 bytes: https://en.wikipedia.org/wiki/MAC_address

leikoilja:
From some [old commit](https://github.com/simon-weber/gmusicapi/commit/30f5080202aaa615ef31b771b1fd0add5b16a315#diff-2bde95b4686ecd42076dd54288ec9957bd7d906f328caade770c20c4ea69c2fcR37) on `gpsoauth` 
image

KapJI:
Why do you call it a mac address then? 🙂
If we can find an example, we can follow its format.

KapJI:
Let's generate 16 chars ID.

leikoilja:
It was some one the stuff reused from `https://gist.github.com/rithvikvibhu/952f83ea656c6782fbd0f1645059055d` calling it mac

leikoilja (approved):

PR #85, opened 2021-04-15
Setup codespell

KapJI:
Codespell can find typos without being annoying.

Also I noticed that prettier was missing. Add some [order](https://www.youtube.com/watch?v=yHE1uM_NF4Y).


leikoilja (approved):
Big thanks

PR #86, opened 2021-04-15
Release 0.3.1

KapJI:
I tested that it works fine in HA, including login flow.

Please create a release and publish the new version after merging 🙂

Btw, I can try to configure poetry to detect version from github release so these version bumps won't be needed. We can even automate publishing to pypi after that. So just creating a github release should do all the work. Do you want to try that? 🙂

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "glocaltokens"
-version = "0.3.0"
+version = "0.3.1"
 authors = ["Ilja Leiko <leikoilja@gmail.com>"]
 classifiers = [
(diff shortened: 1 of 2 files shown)


leikoilja:
@KapJI thanks for trying the package after changes, especially with HA.

The `poetry` automatic bump up would be amazing. 
I could totally see how that could be an amazing custom github action where in configurations you set what files to change when github release PR is open... that could be super useful, so one can set it up in `HA` to bump up `manifest` file as well. I dunno if that would be hard to make thou... a bit of `regex` maybe grasping the specified files? 🤔

leikoilja (approved):
Thank you for the bump!

PR #89, opened 2021-04-16
Action to automatically publish new version on PyPI

### LINKED ISSUE #88: Setup automatic release flow
The plan is to setup action which runs when release in published which will build and upload new package to PyPI. Like here:
https://nicholasnadeau.me/post/2020/8/one-version-to-rule-them-all-keeping-your-python-package-version-number-in-sync-with-git-and-poetry/

The version should be inferred automatically from the latest tag.

@leikoilja can you add `PYPI_TOKEN` to github secrets? Set scope to only this project: https://pypi.org/help/#apitoken

KapJI:
This action is supposed to automatically assign version based on git tag, build package and publish it on PyPI.

Haven't tested everything together, especially the last command.

Locally this will build version `0.0.0` but this should be fine because with this we only need to publish it with this action.

Poetry just merged support of plugins which will be released in v1.2. After that we can use this as a real plugin, right now it's not very simple to install: https://github.com/mtkennerly/poetry-dynamic-versioning

Closes #88

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -35,4 +35,6 @@ jobs:
         run: poetry build
 
       - name: Publish to PyPI
-        run: POETRY_PYPI_TOKEN_PYPI="${{ secrets.PYPI_TOKEN }}" poetry publish
+        run: poetry publish
+        env:
+          POETRY_PYPI_TOKEN_PYPI: ${{ secrets.PYPI_TOKEN }}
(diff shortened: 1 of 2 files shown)


leikoilja (approved):
THIS IS GOING TO BE EPIC! 💯 
Big thanks, @KapJI for looking into it 💥

leikoilja:
Excited to try it! 🚀

KapJI:
@leikoilja let's build 0.3.2 although there won't be any real changes 🙂

PR #92, opened 2021-04-19
Modernise type hints

KapJI:
Same as https://github.com/leikoilja/ha-google-home/pull/218

Under the hood it runs [pyupgrade](https://github.com/asottile/pyupgrade) to update syntax to Python 3.8 standards. So there are also few non typing related changes.


leikoilja (approved):
Gracias!

PR #108, opened 2021-05-14
Update flake8-simplify

KapJI:
New `SIM119` was added which doesn't actually check if constructor is simple enough to be replaced with `dataclass`. Let's disable it.


leikoilja (approved):
Nice! Thanks

PR #145, opened 2021-08-13
Grabbing unique device id from foyer API

### LINKED ISSUE #75: Device unique id parameter
Let's find a way to extract a unique ID from homegraph or `eureka` endpoint to pass it for each of devices

leikoilja:
Discovered that foyer API that we call already contains `device_info` with `device_id` that we can grab and use... How cool? 

Closes #75

### DIFF
--- a/tests/factory/providers.py
+++ b/tests/factory/providers.py
@@ -13,7 +13,7 @@ from glocaltokens.const import (
 from glocaltokens.google.internal.home.foyer.v1_pb2 import GetHomeGraphResponse
 from glocaltokens.utils.token import generate as generate_token
 
-fake = Faker()
+faker = Faker()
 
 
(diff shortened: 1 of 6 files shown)


tests/factory/providers.py
```diff
@@ -13,15 +13,15 @@
 from glocaltokens.google.internal.home.foyer.v1_pb2 import GetHomeGraphResponse
 from glocaltokens.utils.token import generate as generate_token
 
-fake = Faker()
```
ArnyminerZ, line 16: Good fix jeje

leikoilja:
@ArnyminerZ, can you please locally check by running `example/get_tokens.py` script if it returns unique ids for you too? I have checked it locally, but an extra check would be nice. Also, if works - please approve this PR, can't merge it without :D

PR #155, opened 2021-09-02
Update v1.proto

KapJI:
- Define `unique_id` param which I'm going to use.
- Rename `ProjectInfo` to `AgentInfo` to better reflect underlying data and match [documentation](https://developers.google.com/assistant/smarthome/reference/rpc/google.home.graph.v1#agentotherdeviceid) a bit.
- Define empty `GetAssistantRoutinesRequest` message because `None` is not valid.
- Regenerate all files to eliminate all manual changes.

Commands to generate these files:
```
$ protoc --plugin=protoc-gen-grpc_python=~/github/grpc/cmake/build/grpc_python_plugin --grpc_python_out=. --python_out=. --mypy_out=readable_stubs:. --mypy_grpc_out=readable_stubs:. glocaltokens/google/internal/home/foyer/v1.proto
$ git commit
$ pre-commit run --hook-stage manual python-typing-update --all-files
```

Tested locally on my HA instance.


leikoilja (approved):
Pffff, @KapJI, this is incredible! 🤯 
These files were a struggle to understand and generate the first time around when i used them as a 'proof of concept' and they just stayed as is. Haven't had time or heart to dive back into those to try updating them :) Big time thanks for taking a look and implementing that! ❤️ 
And it looks amazing to see that we'll be able to use `unique_id` from the `AgentInfo` for devices unique names. Legend 💥

PR #156, opened 2021-09-03
Unconstrain zeroconf version

KapJI:
This was added by dependabot in https://github.com/leikoilja/glocaltokens/pull/60 with no particular reason.

HA Core uses the latest 0.36.2:

https://github.com/home-assistant/core/blob/70338da50efd6f5537fb8463ac42450bf7010e19/homeassistant/package_constraints.txt#L37

Tested locally that it still works.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -832,7 +832,7 @@ python-versions = "*"
 
 [[package]]
 name = "zeroconf"
-version = "0.29.0"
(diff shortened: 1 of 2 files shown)


leikoilja (approved):
hmm, weird why did dependabot restricted it. Good to know so we could keen an eye for future dependabot actions.

Thanks

PR #158, opened 2021-09-03
Refactor scanner

KapJI:
## Breaking changes
- Rename `google_device` to `network_device`.
- Remove `ip_address` and `port` fields from `Device` constructor, they're only used in tests and add code complexity.
## Scanner changes
- Stop discovery after the timeout. Inspired by pychromecast. Otherwise it can modify list of discovered devices while it's being processed. Also it spams logs.
- Store devices in dict to filter duplicates. Before `update_service` was inserting copies in devices list.
- Handle remove `remove_service`.
## Other changes
- Replace `GoogleDevice` class with `NetworkDevice` named tuple. It should really just hold the data, all checks are moved outside.
- Remove unnecessary type hints from `Final`, they can be inferred automatically.
- Remove few tests which are no longer needed.
- Update some comments.

Tested locally on my HA instance.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -20,12 +20,13 @@ from glocaltokens.const import (
     ANDROID_ID_LENGTH,
     HOMEGRAPH_DURATION,
     JSON_KEY_DEVICE_NAME,
-    JSON_KEY_GOOGLE_DEVICE,
     JSON_KEY_HARDWARE,
     JSON_KEY_IP,
     JSON_KEY_LOCAL_AUTH_TOKEN,
+    JSON_KEY_NETWORK_DEVICE,
     JSON_KEY_PORT,
 )
+from glocaltokens.scanner import NetworkDevice
 from tests.assertions import DeviceAssertions, TypeAssertions
 from tests.factory.providers import HomegraphProvider, TokenProvider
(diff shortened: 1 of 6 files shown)


KapJI:
Jobs failed because of https://github.com/pypa/setuptools_scm/issues/608.

tests/test_client.py
```diff
@@ -382,31 +390,12 @@ def test_initialization__valid(self) -> None:
...
-
-        # With only port
-        device = Device(
-            device_id=faker.uuid4(),
-            device_name=faker.word(),
-            local_auth_token=faker.local_auth_token(),
-            port=faker.port_number(),
-        )
-        self.assertEqual(m_log.call_count, 2)
-        self.assertIsNone(device.local_auth_token)
```
leikoilja, line 403: Since validation got moved to the `scanner` (which by the way is the exact place they should belong to), it would be nice to keep these tests in `test_scanner` to make sure we log any errors happening during device discovery.

leikoilja (approved):
Bud, amazing job! Thanks for a good refactoring. Make sense to move the `ip` and `port` related things into the scanner and extract it as a `network_device`. 🔥 💯

PR #159, opened 2021-09-04
Fix potential infinite loop in get_homegraph

KapJI:
This probably doesn't happen but better safe than sorry.

Retry for 3 times.


leikoilja (approved):
Really good. It would be nice to introduce a small unit test for this method as well, just to be sure we cover the infinite loop case, which might actually be quite a crucial misbehaviour

PR #160, opened 2021-09-04
Match devices by unique ID

KapJI:
Finally use these new IDs.

## Problem
I have a speaker stereo pair which shows in the app (and in home graph) as a single virtual device with model `Google Cast Group`.

But when the main speaker in this pair is not reachable (due of https://github.com/leikoilja/ha-google-home/issues/202), it starts to appear as name of that speaker instead of name of the pair.

This leads to non consistent results from glocaltokens.

## Solution
Local auth tokens are actually per physical device, so glocaltokens should ignore any virtual groups and return only list of actual devices.

This unique id is the same for speaker pair device in home graph and actual device, so in both cases it'll be matched to corresponding network device. Use the name of that network device for consistency.

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -396,8 +396,13 @@ class GLocalAuthenticationTokens:
 
                 network_device = None
                 if network_devices:
-                    LOGGER.debug("Looking for '%s' in local network", item.device_name)
-                    network_device = find_device(item.device_info.agent_info.unique_id)
+                    unique_id = item.device_info.agent_info.unique_id
+                    LOGGER.debug(
+                        "Looking for '%s' (id=%s) in local network",
+                        item.device_name,
+                        unique_id,
+                    )
+                    network_device = find_device(unique_id)
 
                 device = Device(
                     device_id=item.device_info.device_id,
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -377,9 +377,9 @@ class GLocalAuthenticationTokens:
                 logging_level=self.logging_level,
             )
 
-        def find_device(name: str) -> NetworkDevice | None:
+        def find_device(unique_id: str) -> NetworkDevice | None:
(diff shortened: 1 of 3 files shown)


glocaltokens/client.py
```diff
@@ -397,11 +397,13 @@ def find_device(name: str) -> NetworkDevice | None:
                 network_device = None
                 if network_devices:
                     LOGGER.debug("Looking for '%s' in local network", item.device_name)
```
leikoilja, line 399:
```suggestion
                    LOGGER.debug("Looking for '%s'(id=%s) in local network", item.device_name, item.unique_id)
```

glocaltokens/client.py
```diff
@@ -397,11 +397,13 @@ def find_device(name: str) -> NetworkDevice | None:
...
-                    network_device = find_device(item.device_name)
+                    network_device = find_device(item.device_info.agent_info.unique_id)
 
                 device = Device(
                     device_id=item.device_info.device_id,
-                    device_name=item.device_name,
+                    device_name=network_device.name
+                    if network_device is not None
```
leikoilja, line 405:
```suggestion
                    if network_device
```

leikoilja (approved):
Great improvements, thank you, Sir 💯

glocaltokens/client.py
KapJI: I'd follow [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations):

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.
> 
> Also, beware of writing if x when you really mean if x is not None -- e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

leikoilja: uuu interesting. thanks for the read

PR #162, opened 2021-09-06
Small cleanup in proto file

KapJI:
Cleanup `None` message, it was replaced with `GetAssistantRoutinesRequest`. You can't have name `None` in python, it's reserved for constant.

Move high level comment to the top, otherwise it's merged with the next comment by `protoc`.


leikoilja (approved):
Great stuff.

Do you think we should put comments from https://github.com/leikoilja/glocaltokens/pull/155#issue-726372818 to the README to keep your comment up and don't loose that valuable info?

leikoilja:
Thanks, man 🚀

PR #163, opened 2021-09-08
Make generated protobuf files compatible with googlewifi

KapJI:
This should be a short term solution for https://github.com/leikoilja/ha-google-home/issues/330

Generate `v1_pb2.py` from `glocaltokens` folder to avoid name clash.

I tested that googlewifi integration is working now on Home Assistant OS where python implementation of protobuf is used. On my main box with C++ implementation it still doesn't work, but it didn't before, so I guess it's fine.


leikoilja (approved):
Nice, thank you man! 💥 
How about we let the issue author install `master` branch of the integration to check if this fixes the compatibility issue with `googlewifi` and if it solves it we include this fix in release?

KapJI:
Well, I tested on my VM that the issue has been fixed by that. And we need to update glocaltokens version in HA integration anyway for that.

PR #202, opened 2021-12-01
Add possibility of using an address list instead

DurgNomis-drol:
This changes `disable_discovery` from a bool to `address_list` containing a dict. Example dict: ``{"Living room": "192.168.0.69"}` Not setting the address_list results in discovery been enabled. Setting it to and empty dict results in discovery being disabled.

I was not 100% sure on how to best implement this, so please provide feedback if something can be done better. 😄 

Code is tested, but can someone double check to make sure that we are not breaking anything with this.

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -382,7 +382,7 @@ class GLocalAuthenticationTokens:
             # We need to disable flake8-use-fstring because of the brackets,
             # it causes a false positive.
             LOGGER.error(
-                "Invalid dictionary structure for discovery_address_dict "
+                "Invalid dictionary structure for addresses dictionary "
                 "argument. Correct structure is {'device_name': 'ipaddress'}"  # noqa
             )
             return devices
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -372,7 +372,7 @@ class GLocalAuthenticationTokens:
         devices: list[Device] = []
 
         def is_dict_with_valid_ipv4_addresses(data: dict[str, str]) -> bool:
-            # Test if the data structure is correct and if each entry contains a
+            # Validate the data structure is correct and that each entry contains a
             # valid IPv4 address.
             return isinstance(data, dict) and all(
                 isinstance(x, str) and is_valid_ipv4_address(x) for x in data.values()
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -336,7 +336,8 @@ class GLocalAuthenticationTokens:
     def get_google_devices(
         self,
         models_list: list[str] | None = None,
-        discovery_address_dict: dict[str, str] | None = None,
+        disable_discovery: bool = False,
+        addresses: dict[str, str] | None = None,
         zeroconf_instance: Zeroconf | None = None,
         force_homegraph_reload: bool = False,
         discovery_timeout: int = DISCOVERY_TIMEOUT,
@@ -346,7 +347,9 @@ class GLocalAuthenticationTokens:
         and IP and ports if set in models_list.
 
         models_list: The list of accepted model names.
-        discovery_address_dict: Dict of network devices from the local network
+        disable_discovery: Whether or not the device's IP and port should
+          be searched for in the network.
+        addresses: Dict of network devices from the local network
           ({"name": "ip_address"}). If set to `None` will try to automatically
           discover network devices. Disable discovery by setting to `{}`.
         zeroconf_instance: If you already have an initialized zeroconf instance,
@@ -375,10 +378,8 @@ class GLocalAuthenticationTokens:
                 isinstance(x, str) and is_valid_ipv4_address(x) for x in data.values()
             )
 
-        if discovery_address_dict and not is_dict_with_valid_ipv4_addresses(
-            discovery_address_dict
-        ):
-            # We need to disable flake8-use-fstring, because of the brackets,
+        if addresses and not is_dict_with_valid_ipv4_addresses(addresses):
+            # We need to disable flake8-use-fstring because of the brackets,
             # it causes a false positive.
             LOGGER.error(
                 "Invalid dictionary structure for discovery_address_dict "
@@ -391,7 +392,7 @@ class GLocalAuthenticationTokens:
             return devices
 
         network_devices: list[NetworkDevice] = []
-        if discovery_address_dict is None:
+        if disable_discovery is False:
             LOGGER.debug("Automatically discovering network devices...")
             network_devices = discover_devices(
                 models_list,
@@ -406,7 +407,7 @@ class GLocalAuthenticationTokens:
                     return device
             return None
 
-        address_dict = discovery_address_dict if discovery_address_dict else {}
+        address_dict = addresses if addresses else {}
 
         LOGGER.debug("Iterating in %d homegraph devices", len(homegraph.home.devices))
         for item in homegraph.home.devices:
@@ -468,7 +469,8 @@ class GLocalAuthenticationTokens:
         self,
         models_list: list[str] | None = None,
         indent: int = 2,
-        discovery_address_dict: dict[str, str] | None = None,
+        disable_discovery: bool = False,
+        addresses: dict[str, str] | None = None,
         zeroconf_instance: Zeroconf | None = None,
         force_homegraph_reload: bool = False,
     ) -> str:
@@ -478,7 +480,9 @@ class GLocalAuthenticationTokens:
 
         models_list: The list of accepted model names.
         indent: The indentation for the json formatting.
-        discovery_address_dict: Dict of network devices from the local network
+        disable_discovery: Whether or not the device's IP and port should
+          be searched for in the network.
+        addresses: Dict of network devices from the local network
           ({"name": "ip_address"}). If set to `None` will try to automatically
           discover network devices. Disable discovery by setting to `{}`.
         zeroconf_instance: If you already have an initialized zeroconf instance,
@@ -488,7 +492,8 @@ class GLocalAuthenticationTokens:
 
         google_devices = self.get_google_devices(
             models_list=models_list,
-            discovery_address_dict=discovery_address_dict,
+            disable_discovery=disable_discovery,
+            addresses=addresses,
             zeroconf_instance=zeroconf_instance,
             force_homegraph_reload=force_homegraph_reload,
         )
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -16,6 +16,7 @@ from .const import (
     ACCESS_TOKEN_DURATION,
     ACCESS_TOKEN_SERVICE,
     ANDROID_ID_LENGTH,
+    DEFAULT_DISCOVERY_PORT,
     DISCOVERY_TIMEOUT,
     GOOGLE_HOME_FOYER_API,
     HOMEGRAPH_DURATION,
@@ -26,6 +27,7 @@ from .scanner import NetworkDevice, discover_devices
 from .types import DeviceDict
 from .utils import network as net_utils, token as token_utils
 from .utils.logs import censor
+from .utils.network import is_valid_ipv4_address
 
 logging.basicConfig(level=logging.ERROR)
 LOGGER = logging.getLogger(__name__)
@@ -334,7 +336,7 @@ class GLocalAuthenticationTokens:
     def get_google_devices(
         self,
         models_list: list[str] | None = None,
-        disable_discovery: bool = False,
+        discovery_address_dict: dict[str, str] | None = None,
         zeroconf_instance: Zeroconf | None = None,
         force_homegraph_reload: bool = False,
         discovery_timeout: int = DISCOVERY_TIMEOUT,
@@ -344,8 +346,9 @@ class GLocalAuthenticationTokens:
         and IP and ports if set in models_list.
 
         models_list: The list of accepted model names.
-        disable_discovery: Whether or not the device's IP and port should
-          be searched for in the network.
+        discovery_address_dict: Dict of network devices from the local network
+          ({"name": "ip_address"}). If set to `None` will try to automatically
+          discover network devices. Disable discovery by setting to `{}`.
         zeroconf_instance: If you already have an initialized zeroconf instance,
           use it here.
         force_homegraph_reload: If the stored homegraph should be generated again.
@@ -365,13 +368,31 @@ class GLocalAuthenticationTokens:
 
         devices: list[Device] = []
 
+        def is_dict_with_valid_ipv4_addresses(data: dict[str, str]) -> bool:
+            # Test if the data structure is correct and if each entry contains a
+            # valid IPv4 address.
+            return isinstance(data, dict) and all(
+                isinstance(x, str) and is_valid_ipv4_address(x) for x in data.values()
+            )
+
+        if discovery_address_dict and not is_dict_with_valid_ipv4_addresses(
+            discovery_address_dict
+        ):
+            # We need to disable flake8-use-fstring, because of the brackets,
+            # it causes a false positive.
+            LOGGER.error(
+                "Invalid dictionary structure for discovery_address_dict "
+                "argument. Correct structure is {'device_name': 'ipaddress'}"  # noqa
+            )
+            return devices
+
         if homegraph is None:
             LOGGER.debug("Failed to fetch homegraph")
             return devices
 
         network_devices: list[NetworkDevice] = []
-        if disable_discovery is False:
-            LOGGER.debug("Getting network devices...")
+        if discovery_address_dict is None:
+            LOGGER.debug("Automatically discovering network devices...")
             network_devices = discover_devices(
                 models_list,
                 timeout=discovery_timeout,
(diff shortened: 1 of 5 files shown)


DurgNomis-drol:
I am unsure of why the linter fails to install poetry

glocaltokens/client.py
```diff
@@ -334,7 +335,7 @@ def get_homegraph(self, auth_attempts: int = 3) -> GetHomeGraphResponse | None:
     def get_google_devices(
         self,
         models_list: list[str] | None = None,
-        disable_discovery: bool = False,
+        address_list: Dict[str, str] | None = None,
```
leikoilja, line 338: What do you think about naming it `discovery_address_list`?

glocaltokens/client.py
```diff
@@ -370,7 +371,7 @@ def get_google_devices(
             return devices
 
         network_devices: list[NetworkDevice] = []
-        if disable_discovery is False:
+        if address_list is None:
             LOGGER.debug("Getting network devices...")
```
leikoilja, line 375: Can we please adjust the log message here? 
```suggestion
        if address_list is None:
            LOGGER.debug("Automatically discovering network devices...")
```

glocaltokens/client.py
```diff
@@ -344,7 +345,7 @@ def get_google_devices(
         and IP and ports if set in models_list.
 
         models_list: The list of accepted model names.
-        disable_discovery: Whether or not the device's IP and port should
+        address_list: Whether or not the device's IP and port should
           be searched for in the network.
```
leikoilja, line 349:
```suggestion
        address_list: Dict of network devices from the local network ({"name": "ip_address"}). If set to `None` will try to automatically discover network devices. Disable discovery by setting to `{}`
```

leikoilja: Also it's not really a `list` per say. Maybe we can call it `discovery_addresses` to avoid confusion of `list -> []`

glocaltokens/client.py
```diff
@@ -405,14 +409,30 @@ def find_device(unique_id: str) -> NetworkDevice | None:
                         unique_id,
                     )
                     network_device = find_device(unique_id)
+                else:
```
leikoilja, line 412: I suggest instead of `else` we do smth like:
```python
elif item.device_name in address_list:
  network_device = NetworkDevice(
                        name=item.device_name,
                        ip_address=address_list[item.device_name],
                        port=0,
                        model=item.hardware.model,
                        unique_id=item.device_info.device_id,
                    )
```
But i m a bit unsure why do we set post to `0`? Can you please check what port is here for auto discovered network items? It's probably some sort of specific discovery port. If so, we need to add it to `const` and use as static

glocaltokens/client.py
```diff
@@ -405,14 +409,30 @@ def find_device(unique_id: str) -> NetworkDevice | None:
...
                 device = Device(
                     device_id=item.device_info.device_id,
                     device_name=network_device.name
                     if network_device is not None
                     else item.device_name,
                     local_auth_token=item.local_auth_token,
-                    network_device=network_device,
+                    network_device=network_device
+                    if network_device is not None
+                    else man_disc_device,
```
leikoilja, line 435: We can have it a bit cleaner if we choose to go with `elif` as suggested above
```suggestion
                    network_device=network_device,
```

tests/factory/providers.py
```diff
@@ -43,14 +43,19 @@ def local_auth_token(self) -> str:
 class HomegraphProvider(TokenProvider):
     """Homegraph provider"""
 
-    def homegraph_device(self) -> GetHomeGraphResponse.Home.Device:
+    def homegraph_device(
+        self, device_name: str | None = None
+    ) -> GetHomeGraphResponse.Home.Device:
         """Using the content from test requests as reference"""
+        if device_name is None:
+            device_name = faker.word()
```
leikoilja, line 51:
```suggestion
            device_name = device_name if device_name else faker.word()
```

tests/test_client.py
```diff
@@ -322,19 +322,22 @@ def test_get_homegraph_retries(
...
+        fake_device_name = faker.word()
+        homegraph_device = faker.homegraph_device(device_name=fake_device_name)
         m_get_homegraph.return_value.home.devices = [homegraph_device]
 
         # With no discover_devices, with no model_list
-        google_devices = self.client.get_google_devices(disable_discovery=True)
+        google_devices = self.client.get_google_devices(
+            address_list={fake_device_name: "1.1.1.1"}
+        )
         self.assertEqual(len(google_devices), 1)
```
leikoilja, line 333: It would be nice to fake generate ip address using `faker` here and then assert that `google_device` has the name same as `fake_device_name` and ip is the same as you fake generated ip

leikoilja (changes requested):
That's an amazing improvements, thanks man! Have a few suggestions

glocaltokens/client.py
DurgNomis-drol: We discover the port through Zeroconf. The reason I set it to `0` is because i didn't want to mess with `NetworkDevice` and change the typing for port to `int | None` instead of just `int`. Should i just make a `const` with that standard port it returns or is it better that we return something like `None` or `0` to signal that the port was not discovered?

glocaltokens/client.py
DurgNomis-drol: Fine with me, i struggled to find a good name 😃

glocaltokens/client.py
DurgNomis-drol: You are right, what about `discovery_address_dict`?

glocaltokens/client.py
DurgNomis-drol: See below, have named it `discovery_address_dict`

glocaltokens/client.py
DurgNomis-drol: Partly resolved this, need you input about the port thing 😄

glocaltokens/client.py
DurgNomis-drol: Changed it to `discovery_address_dict`

DurgNomis-drol:
Please see my changes and the last comment that are open 😃

glocaltokens/client.py
```diff
@@ -405,6 +411,15 @@ def find_device(unique_id: str) -> NetworkDevice | None:
                         unique_id,
                     )
                     network_device = find_device(unique_id)
+                elif item.device_name in discovery_address_dict:
+                    network_device = NetworkDevice(
+                        name=item.device_name,
+                        ip_address=discovery_address_dict[item.device_name]
+                        or "0.0.0.0",
```
leikoilja, line 418: I still believe we should not default the IP to 0.0.0.0.
According to wikipedia

> the address 0.0. 0.0 is a non-routable meta-address used to designate an invalid, unknown or non-applicable target

We are getting `discovery_address_dict` from the user so it must contain name and IP of devices, right? If you want to make it extra rigid we can add an inout validation to the `get_google_devices` function to check the data in `discovery_address_dict` to contain expected name/ip pairs, so if the input `dict` is invalid we raise the error :)

```suggestion
                        ip_address=discovery_address_dict[item.device_name],
```

glocaltokens/client.py
```diff
@@ -405,6 +411,15 @@ def find_device(unique_id: str) -> NetworkDevice | None:
...
+                        port=0,
```
leikoilja, line 419: hmm, i m not so sure what to do with this port. Perhaps you are right and we should not assign autodiscovery port if the device was matched by manual entry. Shall we maybe default to port `80` then, which is the default access port? :) But either way, let's make it a static and move to `const` file

tests/test_client.py
```diff
@@ -322,19 +322,26 @@ def test_get_homegraph_retries(
...
+        google_devices = self.client.get_google_devices(
+            discovery_address_dict={fake_device_name: fake_ip_address}
+        )
         self.assertEqual(len(google_devices), 1)
 
         google_device = google_devices[0]
         self.assertDevice(google_device, homegraph_device)
+        self.assertIsNotNone(google_device.network_device)
+        if google_device.network_device is not None:
+            self.assertEqual(google_device.network_device.ip_address, fake_ip_address)
```
leikoilja, line 340: Flip yeh, let's make those tests robust! 💪

leikoilja (approved):
Yay! Amazing job, man! Thanks 💯 
Just a minor comment about default IPs/port and it's good to go 🚀

glocaltokens/client.py
DurgNomis-drol: I think it is a great idea to validate the input

glocaltokens/client.py
DurgNomis-drol: I will move it to a const, don't think it is a good idea to set it to port 80 though.

glocaltokens/client.py
DurgNomis-drol: I have added input validation to the dict, it also checks if the IP address is a valid IPv4 address 😄

DurgNomis-drol:
I have no idea of what went wrong and why it essentially deletes and the adds `client.py` when the new lines again, but i don't have time to fixed it today, will fixed it tomorrow

DurgNomis-drol:
Can it be line endings?

DurgNomis-drol:
Got it fixed, i have no idea what went wrong but rolled back to a working commit and fixed the linter.

@leikoilja Can I get you to release a new version after this is merged? :smiley:

poetry.lock
```diff
@@ -45,17 +45,17 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "attrs"
-version = "20.3.0"
```
KapJI, line 48: Looks like you accidentally ran `poetry upgrade`, please revert this file. If you want to upgrade dependencies, create a separate PR.

DurgNomis-drol: Will revert it, but i didn't run `poetry upgrade`.

KapJI:
Can you give some context about what problem you're solving with this?

poetry.lock
DurgNomis-drol: Can it be that i ran `poetry update` on this branch?

DurgNomis-drol:
> Can you give some context about what problem you're solving with this?

https://github.com/leikoilja/ha-google-home/issues/340 is just one of them.

As i am working on moving the ha-google-home code to the separate module, i ran into some problems with zeroconf on WSL2. Code run in WSL2 can't discover devices through zeroconf. This just makes it much easier for me to test it when i am developing.

poetry.lock
DurgNomis-drol: Should be fixed now

KapJI:
Thanks, makes sense.

I think breaking compatibility for such change is not necessary, so let's keep `disable_discovery`.

Instead add `extra_addresses` argument in the end. IP address should be the key. And append it to list of discovered devices.

DurgNomis-drol:
@KapJI No. We need to pass the name of the device as the key and the IP address as the value. If we don't, we can't match the IP address to the device and so on.

Making it a separate argument, means the user now needs to set `disable_discovery` to `true` and then set a dict with addresses. Using my way, we just disable it automatically if we have dict containing addresses.

KapJI:
You don't need to disable discovery completely. If it crashes on WSL2, you can disable it automatically if running inside WSL2.

Mapping from IP to name makes more sense from UX perspective which I believe should come before implementation. You can build reverse dict here if needed.

DurgNomis-drol:
I Think you are missing the point of what i am trying to do with this PR.

It dosen't make sense to reverse it like you are takling about. We match on device name not the IP.

KapJI:
Ok, having name as a key is fine actually.

leikoilja:
Nice!
Awesome thing we have more opinions and reviewers in here. I am not particularly worried about breaking compatibility, since we don't rely on that in `ha-google-home` and we can't really know if there is anyone using `disable_discovery` kwarg at all, so IMHO it shouldn't be a biggie. I ll leave it up to you, @DurgNomis-drol, to decide whether or not you want to refactor it into a separate kwarg keeping `disable_discovery`, but feel free to merge when you ready! 🚀 
p.s. also given you the full repo permissions in the repo, not sure why it wasn't set in the first place 💯 
Thanks man! 🔥

DurgNomis-drol:
@leikoilja Thanks for both invites! Haven really been active here, so that is fine that i am only getting it now :laughing: 

I will let @KapJI suggestions about `disable_discovery` tumble around in my head until tomorrow, just to make sure i am not missing something :smile: Right now i can't see the use for a separate argument

KapJI:
Aoproach I proposes should be more flexible. Basically you can still use discovery and extend it with your own list. Also empty dict for disabling discovery while None enables it, is not very clear API.

DurgNomis-drol:
@KapJI Good point about it not being clear the way i have proposed.

glocaltokens/client.py
```diff
@@ -365,13 +371,29 @@ def get_google_devices(
 
         devices: list[Device] = []
 
+        def is_dict_with_valid_ipv4_addresses(data: dict[str, str]) -> bool:
+            # Test if the data structure is correct and if each entry contains a
```
leikoilja, line 375:
```suggestion
            # Validate the data structure is correct and that each entry contains a
```

glocaltokens/client.py
```diff
@@ -365,13 +371,29 @@ def get_google_devices(
...
+            # valid IPv4 address.
+            return isinstance(data, dict) and all(
+                isinstance(x, str) and is_valid_ipv4_address(x) for x in data.values()
+            )
+
+        if addresses and not is_dict_with_valid_ipv4_addresses(addresses):
+            # We need to disable flake8-use-fstring because of the brackets,
+            # it causes a false positive.
+            LOGGER.error(
+                "Invalid dictionary structure for discovery_address_dict "
```
leikoilja, line 385:
```suggestion
                "Invalid dictionary structure for addresses dictionary "
```

glocaltokens/client.py
```diff
@@ -365,13 +371,29 @@ def get_google_devices(
...
         devices: list[Device] = []
 
+        def is_dict_with_valid_ipv4_addresses(data: dict[str, str]) -> bool:
+            # Test if the data structure is correct and if each entry contains a
+            # valid IPv4 address.
+            return isinstance(data, dict) and all(
+                isinstance(x, str) and is_valid_ipv4_address(x) for x in data.values()
+            )
+
+        if addresses and not is_dict_with_valid_ipv4_addresses(addresses):
```
leikoilja, line 381: Tbh i don't know if there is any benefit of the proposed approach but it looks more readable, isn't it? Or is it just me? :D 

```suggestion
        if addresses and is_dict_with_valid_ipv4_addresses(addresses) is False:
```

glocaltokens/client.py
```diff
@@ -385,6 +407,8 @@ def find_device(unique_id: str) -> NetworkDevice | None:
                     return device
             return None
 
+        address_dict = addresses if addresses else {}
```
leikoilja, line 410: This could probably just be `addresses = addresses if addresses else {}`, no need for a new variable?

glocaltokens/client.py
DurgNomis-drol: Your way is more readable, but the other way is the pythonic way to do it, as far as i know

glocaltokens/client.py
DurgNomis-drol: mypy will throw a typing error, because `addresses` is optional `dict`, and when we access it below when setting a `network_devices` we get a typing error expecting `str`, got optional `str`.

glocaltokens/client.py
leikoilja: haha same :D

glocaltokens/client.py
leikoilja: nice to know, thanks 👍

leikoilja:
Amazing work, thank you, @DurgNomis-drol 💯 🔥

leikoilja:
ping @DurgNomis-drol for v0.6 release

glocaltokens/client.py
```diff
@@ -405,6 +429,14 @@ def find_device(unique_id: str) -> NetworkDevice | None:
                         unique_id,
                     )
                     network_device = find_device(unique_id)
+                elif item.device_name in address_dict:
```
KapJI, line 432: This means if discovery is enabled, `address_dict` won't be used. That contradicts the idea to have separate `disable_discovery` flag.

glocaltokens/client.py
```diff
@@ -405,6 +429,14 @@ def find_device(unique_id: str) -> NetworkDevice | None:
...
+                    network_device = NetworkDevice(
+                        name=item.device_name,
+                        ip_address=address_dict[item.device_name],
```
KapJI, line 435: Case when there is no item with such name is not handled which will cause whole integration to crash.

glocaltokens/client.py
KapJI: Also if there's no item with such name in `address_dict`, it's worth to print a warning message.

glocaltokens/client.py
```diff
@@ -335,6 +337,7 @@ def get_google_devices(
         self,
         models_list: list[str] | None = None,
         disable_discovery: bool = False,
+        addresses: dict[str, str] | None = None,
```
KapJI, line 340: It's better to add new arguments in the end, to not break reverse compatibility.

PR #216, opened 2022-02-04
Move grpcio-tools to dev dependencies

KapJI:
I tested locally that with grpcio-tools uninstalled, the integration still works.


leikoilja:
thanks bud! 🔥

PR #219, opened 2022-02-24
Resolving HomeAssistant dependency conflicts

leikoilja:
With some Frenk's [help](https://github.com/home-assistant/wheels-custom-integrations/issues/590#issuecomment-1049845684):

What we had:
```
[tool.poetry.dependencies]
python = "^3.8"
gpsoauth = "^1.0.0"
requests = "^2.25.1"
simplejson = "^3.17.2"
zeroconf = "^0.36.2"

# We want to keep grpcio similar to Home Assistant
# https://github.com/home-assistant/core/blob/9ce2e9e8f4e9afaefc4a66e788d503ebe817c327/homeassistant/package_constraints.txt
grpcio = "1.43.0"
```

Those are pretty hard pins for a library.

requests is limited from =>2.25.1 and <2.26.0, while Home Assistant requires [2.27.1](https://github.com/home-assistant/core/blob/dev/requirements.txt#L21)
Zeroconf has similar issues: https://github.com/home-assistant/core/blob/dev/requirements_all.txt#L2536
grpcio is hard pinned, which a library should never do. Home Assistant side: https://github.com/home-assistant/core/blob/dev/homeassistant/package_constraints.txt#L48-L51

So, on the first quick pass, those are definitely conflicting dependencies (and they've always been). Newer Python pip versions have a new resolver that no longer allows for this. And blocks the installation of packages with conflicts.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -24,16 +24,30 @@ python-versions = "!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,>=2.7"
 
 [[package]]
 name = "astroid"
-version = "2.9.0"
+version = "2.9.3"
 description = "An abstract syntax tree for Python with inference support."
 category = "dev"
 optional = false
-python-versions = "~=3.6"
+python-versions = ">=3.6.2"
 
 [package.dependencies]
 lazy-object-proxy = ">=1.4.0"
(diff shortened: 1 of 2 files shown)


KapJI (approved):
Thanks!

PR #231, opened 2022-04-06
Added `ValueError` handler for long passwords

### LINKED ISSUE #229: Long passwords cause login issues
**Describe the bug**
When long passwords are used (100+ characters), which are [not supported by Google](https://support.google.com/a/answer/139399), an issue is thrown. More info in leikoilja/ha-google-home#430

**Partial logs:**
```
  File "/usr/local/lib/python3.9/site-packages/glocaltokens/client.py", line 218, in get_master_token
    res = perform_master_login(
  File "/usr/local/lib/python3.9/site-packages/gpsoauth/__init__.py", line 127, in perform_master_login
    "EncryptedPasswd": google.construct_signature(
  File "/usr/local/lib/python3.9/site-packages/gpsoauth/google.py", line 58, in construct_signature
    encrypted_login = cipher.encrypt((email + "\x00" + password).encode("utf-8"))
  File "/usr/local/lib/python3.9/site-packages/Cryptodome/Cipher/PKCS1_OAEP.py", line 115, in encrypt
    raise ValueError("Plaintext is too long.")
ValueError: Plaintext is too long.
```

**Possible solution:**
Add a try-catch to [/glocaltokens/client.py#L220](https://github.com/leikoilja/glocaltokens/blob/master/glocaltokens/client.py#L220) which catches the `ValueError`.

ArnyminerZ:
Fixes #229

### DIFF
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -212,21 +212,28 @@ class GLocalAuthenticationTokens:
             LOGGER.error("Username and password are not set.")
             return None
 
-        if not self.master_token:
-            LOGGER.debug(
-                "There is no stored master_token, "
-                "logging in using username and password"
+        try:
+            if not self.master_token:
+                LOGGER.debug(
+                    "There is no stored master_token, "
+                    "logging in using username and password"
+                )
+                res = perform_master_login(
+                    self.username, self.password, self.get_android_id()
+                )
+                if "Token" not in res:
+                    LOGGER.error("[!] Could not get master token.")
+                    LOGGER.debug("Request response: %s", res)
+                    return None
+                self.master_token = res["Token"]
+            LOGGER.debug("Master token: %s", censor(self.master_token))
+            return self.master_token
+        except ValueError:
+            LOGGER.error(
+                "A ValueError exception has been thrown, this usually is related to a "
+                "password length that exceeds the boundaries (too long)."
             )
-            res = perform_master_login(
-                self.username, self.password, self.get_android_id()
-            )
-            if "Token" not in res:
-                LOGGER.error("[!] Could not get master token.")
-                LOGGER.debug("Request response: %s", res)
-                return None
-            self.master_token = res["Token"]
-        LOGGER.debug("Master token: %s", censor(self.master_token))
-        return self.master_token
+            return None
 
     def get_access_token(self) -> str | None:
         """Return existing or fetch access_token"""
--- a/glocaltokens/client.py
+++ b/glocaltokens/client.py
@@ -212,28 +212,28 @@ class GLocalAuthenticationTokens:
             LOGGER.error("Username and password are not set.")
             return None
 
-        try:
-            if not self.master_token:
-                LOGGER.debug(
-                    "There is no stored master_token, "
-                    "logging in using username and password"
-                )
+        if not self.master_token:
+            LOGGER.debug(
+                "There is no stored master_token, "
+                "logging in using username and password"
+            )
+            try:
                 res = perform_master_login(
                     self.username, self.password, self.get_android_id()
                 )
-                if "Token" not in res:
-                    LOGGER.error("[!] Could not get master token.")
-                    LOGGER.debug("Request response: %s", res)
-                    return None
-                self.master_token = res["Token"]
-            LOGGER.debug("Master token: %s", censor(self.master_token))
-            return self.master_token
-        except ValueError:
-            LOGGER.error(
-                "A ValueError exception has been thrown, this usually is related to a "
-                "password length that exceeds the boundaries (too long)."
-            )
-            return None
+            except ValueError:
+                LOGGER.error(
+                    "A ValueError exception has been thrown, this usually is related"
+                    "to a password length that exceeds the boundaries (too long)."
+                )
+                res = {}
+            if "Token" not in res:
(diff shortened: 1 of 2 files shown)


ArnyminerZ:
Not sure if this would be useful @leikoilja, but maybe adding something like this to `test_client.py`?

```python3
# For passwords that are too long
m_perform_master_login.side_effect = ValueError("Plaintext is too long.")
self.assertIsNone(self.client.get_master_token())
```

glocaltokens/client.py
```diff
@@ -212,21 +212,28 @@ def get_master_token(self) -> str | None:
...
+                    LOGGER.error("[!] Could not get master token.")
+                    LOGGER.debug("Request response: %s", res)
+                    return None
+                self.master_token = res["Token"]
+            LOGGER.debug("Master token: %s", censor(self.master_token))
+            return self.master_token
+        except ValueError:
+            LOGGER.error(
+                "A ValueError exception has been thrown, this usually is related to a "
+                "password length that exceeds the boundaries (too long)."
```
leikoilja, line 234: No need to wrap the entire logic block in try-except, i suggest we just wrap the API all with it like so (this is pseudo, please don't accept the suggestion :D )
```suggestion
            if not self.master_token:
                LOGGER.debug(
                    "There is no stored master_token, "
                    "logging in using username and password"
                )
                try:
                  res = perform_master_login(
                      self.username, self.password, self.get_android_id()
                  )
                except ValueError as e:
                 LOGGER.error(
                  "A ValueError exception has been thrown, this usually is related to a "
                  "password length that exceeds the boundaries (too long). Error message: %s"
                  % e
                )
                if "Token" not in res:
                    LOGGER.error("[!] Could not get master token.")
                    LOGGER.debug("Request response: %s", res)
                    return None
                self.master_token = res["Token"]
            LOGGER.debug("Master token: %s", censor(self.master_token))
            return self.master_token
           
```

glocaltokens/client.py
```diff
@@ -212,21 +212,28 @@ def get_master_token(self) -> str | None:
...
+                    self.username, self.password, self.get_android_id()
+                )
+                if "Token" not in res:
+                    LOGGER.error("[!] Could not get master token.")
+                    LOGGER.debug("Request response: %s", res)
+                    return None
+                self.master_token = res["Token"]
+            LOGGER.debug("Master token: %s", censor(self.master_token))
+            return self.master_token
+        except ValueError:
```
leikoilja, line 231: Are we sure the google sdk raises ValueError is the password is too long? I think we might need to  manually test it out by passing toooo long password and see what SDK will raise

leikoilja (commented):
Yay, that's amazing, thank you @ArnyminerZ 💪 💥 

And the test you are proposing would also be very needed 💯 
What we could do in that unittest is:
1. pass in a password with char length > 100. 
2. moch `side_effect` to raise `ValueError` just as you do
3. Assert the client returns `None` and we could also assert the `logger` is logging the `error` for us :)

glocaltokens/client.py
ArnyminerZ: Well, the reported error in the HA integration shows that exception. In fact, maybe the Google limitation is related to some kind of constrain on the hashing algorithm, that's why an error is thrown before even being sent to the servers. Whatsoever, I agree that further testing should be done.

ArnyminerZ:
@leikoilja take a look at the tests. They do pass, but not sure if they are correct.

glocaltokens/client.py
```diff
@@ -217,9 +217,16 @@ def get_master_token(self) -> str | None:
...
+            try:
+                res = perform_master_login(
+                    self.username, self.password, self.get_android_id()
+                )
+            except ValueError:
+                LOGGER.error(
+                    "A ValueError exception has been thrown, this usually is related"
+                    "to a password length that exceeds the boundaries (too long)."
+                )
+                res = {}
```
leikoilja, line 229: A small suggestion
```suggestion
            res = {}
            try:
                res = perform_master_login(
                    self.username, self.password, self.get_android_id()
                )
            except ValueError:
                LOGGER.error(
                    "A ValueError exception has been thrown, this usually is related"
                    "to a password length that exceeds the boundaries (too long)."
                )
```

tests/test_client.py
```diff
@@ -186,6 +186,17 @@ def test_get_master_token(
         master_token = self.client.get_master_token()
         self.assertEqual(expected_master_token, master_token)
 
+        # Reset mocks
+        m_perform_master_login.reset_mock()
+        m_log.reset_mock()
+
+        # For passwords that are too long
+        m_perform_master_login.side_effect = ValueError("Plaintext is too long.")
+        self.client.invalidate_master_token()
```
leikoilja, line 195: Let's invalidate master token before resetting `m_log` and then we can check that m_log is just called once and that's when our code is logging and `error`. In reality we should have just patched `logger.error` and asserted that `error` logging level was called.

ArnyminerZ: Well, `m_log` I believe is in fact called twice: first on the try-catch, and then on the `if "Token" not in res` check. Therefore the assertion call count should be `2`. Correct me if I'm wrong

leikoilja (approved):

leikoilja:
Awesome stuff, thank you, @ArnyminerZ 💪 🚀

PR #287, opened 2022-09-02
Relax requirements for deps which are pinned in HA

KapJI:
This should allow minor versions to be updated on HA side without changed in glocaltokens.


leikoilja:
Thank you, @KapJI

PR #337, opened 2023-02-28
Use ghome-foyer-api package

KapJI:
Google Wifi integration finally merged my pull request, so we can migrate as well.

Move generated files to external package.

More context: https://github.com/djtimca/googlewifi-api/pull/7


leikoilja:
that's awesome, thank for moving them out 💪

PR #383, opened 2023-05-10
Pin grpcio to 1.51.1

leikoilja:
@KapJI, why do we allow grpcio to be automatically upgraded? 
This is quite crucial dependency for this package and home assistant still has it pinned to 1.51.1. 
Why would we not pin it as well to eliminate issues with it? (i had issues installing 1.53.0 on devcontainer while testing ha-google-home)


leikoilja:
@KapJI :)

KapJI:
Not pinning it allows to resolve dependencies when HA upgrades it. For each pip invocation from HA they add `package_constraints.txt`. So this allows to always use version which HA uses.

So pinning it here is not necessary and can be harmful.

Our CI depends on `homeassistant-stubs` which depends on `homeassistant` package, so for CI it should be correctly pinned.

Instead we can instruct dependabot to not upgrade these libs in `poetry.lock`.

leikoilja:
but does it not raise inconsistency between this package and HA?
or because HA is requiring `grpcio==1.51.1`, that version would be installed on HA instance and when ha-google-home is pulled it won't be requiring `grpcio` update but would work by using `grpcio==1.51.1`?

KapJI:
The requirements are combined. If HA is requesting `grpcio==1.51.1` and glocaltokens for example `grpcio>=1.51.0` then version 1.51.1 will be used.

If HA changes to `grpcio==1.52.0` but glocaltokens depends on strictly `grpcio==1.51.1` then dependency resolution will fail.

That's why we shouldn't pin dependencies in libraries for better compatibility.

PR #470, opened 2023-11-21
Censor password better

KapJI:
Revealing password length and first letter discloses some unnecessary info.


leikoilja (approved):
nice, thanks

PR #556, opened 2024-09-09
Enable caches on CI

KapJI:
Cache installed dependencies with Poetry and Pre-Commit hooks to make CI much faster.

I setup it in my other repos and it work really well.


leikoilja (commented):
thanks @KapJI 🙌

PR #560, opened 2024-09-12
Migrate from poetry to uv

KapJI:
uv is written in rust and much faster than poetry.

It now can be used as a full featured package manager: https://astral.sh/blog/uv-unified-python-packaging


leikoilja:
Boom what an amazing change 🚀  thank you 🙌

PR #568, opened 2024-11-05
Update dependencies

KapJI:
Bunch of misc changes:
- Update `ghome-foyer-api` dependency to v1.1.1 which adds protobuf v5 support.
- Sync other dependencies with `homeassistant==2024.11.0`.
- Update dev dependencies.
- Remove `pre-commit` from dev dependencies. It should be installed globally by `uv tool install` or run through `uvx`. Running it through `uv` may cause caching issues if some transitive dependency specify version of `uv` dependency. And `homeassistant` [does](https://github.com/home-assistant/core/blob/79901cede985830ab053c8945e253d7b39c61f8e/homeassistant/package_constraints.txt#L65).
- Publish using `uv` without `twine`. `uv` develops very fast and now supports publishing without additional tools.
- Publish without using `PYPI_TOKEN`.
- Migrate to `[dependency-groups]` in `pyproject.toml` which is defined by standard and now supported by `uv`.
- Use official `astral-sh/setup-uv@v3` which implements caching support.
- Install Python with `uv`, it's faster and more flexible. E.g. it checks required python version in `pyproject.toml`.
- Downgrade `crazy-max/ghaction-github-labeler` to v5.0.0, v5.1.0 is broken. Prevent updates with dependabot.
- Add `automerge` label for Kodiak

### DIFF
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -9,21 +9,16 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Setup Python
-      uses: actions/setup-python@v5
-      with:
-        python-version: ${{ inputs.python-version }}
-
     - name: Install uv
-      uses: yezz123/setup-uv@v4
-
-    - name: Cache project dependencies
-      uses: actions/cache@v4
+      uses: astral-sh/setup-uv@v3
       with:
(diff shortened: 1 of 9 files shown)


leikoilja:
nice, thank you 🙌

